PR #57, opened 2020-04-18
removed extra 'that'

imorti:


jswoods:
First that is a conjunction and second that is a demonstrative pronoun.

Issue #92, opened 2020-06-10
Ensure operator can manage multiple clusters


SaaldjorMike:
- [x] Ensure service account names doesn't conflict if the operator manages service accounts and two HumioClusters are deployed in the same namespace: https://github.com/humio/humio-operator/pull/97
- [x] Ensure operator can watch and manage clusters humio cluster CR's in multiple namespaces: https://github.com/humio/humio-operator/pull/98
- [ ] Install a per-`helm install` SCC and update `users` list in the  `SecurityContextConstraint` to include service accounts as needed: https://github.com/humio/humio-operator/pull/99

PR #97, opened 2020-06-11
Refactor naming of k8s resources created for a HumioCluster.

SaaldjorMike:
This moves us one step closer to closing https://github.com/humio/humio-operator/issues/92

This fixes the problems with resource name conflicts when handling
multiple HumioCluster resources in the same namespace.

This breaks the `crc` cluster things for now, because we cannot pre-populate the service account names in the `users` property in the `SecurityContextConstraint` resource, because service accounts are now named based on the CR names. I'm thinking we should just merge this anyway because #92 already contains a step to fix the SCC issues. Let me know if you think the SCC changes needs to be done in this PR, then I'll update it, if not, then I will take a look at this in a followup PR.

### DIFF
--- a/pkg/controller/humiocluster/defaults.go
+++ b/pkg/controller/humiocluster/defaults.go
@@ -10,27 +10,31 @@ import (
 )
 
 const (
-	image                          = "humio/humio-core:1.12.0"
-	targetReplicationFactor        = 2
-	storagePartitionsCount         = 24
-	digestPartitionsCount          = 24
-	nodeCount                      = 3
-	humioPort                      = 8080
-	elasticPort                    = 9200
-	humioServiceAccountName        = "humio-service-account"
-	initServiceAccountName         = "init-service-account"
-	initServiceAccountSecretName   = "init-service-account"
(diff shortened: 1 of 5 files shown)


pkg/controller/humiocluster/defaults.go
```diff
@@ -113,48 +117,60 @@ func humioServiceAccountNameOrDefault(hc *humioClusterv1alpha1.HumioCluster) str
 	if hc.Spec.HumioServiceAccountName != "" {
 		return hc.Spec.HumioServiceAccountName
 	}
-	return humioServiceAccountName
+	return fmt.Sprintf("%s-%s-%s", hc.Namespace, hc.Name, humioServiceAccountNameSuffix)
```
jswoods, line 120: For all of these, is there a reason we include the namespace in the name of the resource? I think all of these resources are already namespaced.

SaaldjorMike: Right! That was actually my initial plan. Doh! I will update the PR. Thanks!

PR #98, opened 2020-06-11
Refactor helm chart to allow watching a single, multiple or all namespaces

SaaldjorMike:
This builds on top of https://github.com/humio/humio-operator/pull/97 and moves us yet another step closer to closing #92

~I was hoping it was as simple as creating a pair of `Role` & `RoleBinding` resources for each namespace we want to watch and then set the `WATCH_NAMESPACE` for the Humio operator to be a comma-separated list, but that doesn't seem like the case. The operator doesn't accept a comma-separated list in `WATCH_NAMESPACE`. Here's an example of the output when trying to pass on multiple namespaces:~

~`
E0611 13:09:04.316750       1 reflector.go:123] pkg/mod/k8s.io/client-go@v0.0.0-20191016111102-bec269661e48/tools/cache/reflector.go:96: Failed to list *v1alpha1.HumioExternalCluster: humioexternalclusters.core.humio.com is forbidden: User "system:serviceaccount:default:humio-operator" cannot list resource "humioexternalclusters" in API group "core.humio.com" in the namespace "default,flafgnyf"
`~

~@jswoods @schofield Do you think we even need the ability to watch multiple namespaces? If not/unknown, I will not pursue this right now.~

Scratch that, the PR now includes support for single, multiple and all namespaces.

### DIFF
--- a/charts/humio-operator/README.md
+++ b/charts/humio-operator/README.md
@@ -64,6 +64,7 @@ Parameter | Description | Default
 `operator.image.repository` | operator container image repository | `humio/humio-operator`
 `operator.image.tag` | operator container image tag | `v0.0.3`
 `operator.rbac.create` | automatically create operator RBAC resources | `true`
+`operator.watchNamespaces` | list of namespaces the operator will watch for resources (if empty, it watches all namespaces) | `[]`
 `installCRDs` | automatically install CRDs. NB: if this is set to true, custom resources will be removed if the Helm chart is uninstalled | `false`
 `openshift` | install additional RBAC resources specific to OpenShift | `false`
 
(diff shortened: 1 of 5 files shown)


charts/humio-operator/README.md
```diff
@@ -64,6 +64,7 @@ Parameter | Description | Default
 `operator.image.repository` | operator container image repository | `humio/humio-operator`
 `operator.image.tag` | operator container image tag | `v0.0.3`
 `operator.rbac.create` | automatically create operator RBAC resources | `true`
+`operator.watchNamespace` | comma-separated list of namespaces the operator will watch for resources (if empty, it watches all namespaces) | `[]`
```
jswoods, line 67: I wonder if this should be `watchNamespaces` since it's plural.

schofield:
And here I was going to say we shouldn't care about multiple namespaces...

PR #109, opened 2020-07-06
Fix race in e2e tests

jswoods:


test/e2e/humiocluster_test.go
```diff
@@ -132,9 +146,14 @@ func HumioClusterWithPVCs(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
+
+	done <- true
+	wg.Wait()
 }
 
-func printKubectlcommands(t *testing.T, namespace string) {
+func printKubectlcommands(t *testing.T, namespace string, wg *sync.WaitGroup, done chan bool) {
```
SaaldjorMike, line 154: If I'm not mistaken we only read from this channel within this function/routine, so might as well explicitly state that:
```suggestion
func printKubectlcommands(t *testing.T, namespace string, wg *sync.WaitGroup, done <-chan bool) {
```

PR #118, opened 2020-07-14
Add helm chart migration guide

jswoods:


docs/migration/README.md
```diff
@@ -0,0 +1,830 @@
...
+        secretKeyRef:
+          name: humio-google-oauth-secret
+          key: supersecretkey
+    - name: GOOGLE_OAUTH_CLIENT_ID
+      value: YOURCLIENTID
+```
+
+#### HumioCluster
+```yaml
+spec:
```
SaaldjorMike, line 642: I think we're missing the part about specifying the `IdpCertificateSecretName` on the HumioCluster? From what I can see the old helm chart assumed the secret holding the IDP certificate was called `idp-certificate` (see https://github.com/humio/humio-helm-charts/blob/69556adbcc2cc24bab4ce51c2a307653526fcb3d/charts/humio-core/templates/statefulset.yaml#L297-L301), whereas the operator by default assumes the secret is prefixed with the HumioCluster name which would be something like `examplecluster-idp-certificate` (see https://github.com/humio/humio-operator/blob/f087180c556d396004542e87850c52808f504345/pkg/controller/humiocluster/defaults.go#L167-L172).

docs/migration/README.md
```diff
@@ -0,0 +1,830 @@
...
+      value: "$(ZOOKEEPER_URL)"
+    - name: LOCAL_STORAGE_PERCENTAGE
+      value: "80"
+    - name: LOCAL_STORAGE_MIN_AGE_DAYS
+      value: "7"
+```
+
+#### HumioCluster
+
+Note that the Humio Operator does not currently support GCP bucket storage. See issue https://github.com/humio/humio-operator/issues/117.
```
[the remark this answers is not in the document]
jswoods: Oops, I guess I missed updating this with the gcp change. Good catch

docs/migration/README.md
jswoods: I think this should work by default since we call that same method when creating the pod: https://github.com/humio/humio-operator/blob/f087180c556d396004542e87850c52808f504345/pkg/controller/humiocluster/pods.go#L350

SaaldjorMike: Right, but we what I am missing here is instructions that tells the user to create the secret with the name matching what that function returns by default. The name for the secret in the old helm chart is not the same as what we assume the name is for the HumioCluster's.

SaaldjorMike: The old helm charts mounted in the k8s secret named `idp-certificate` whereas the operator assumes that secret is `fmt.Sprintf("%s-%s", hc.Name, idpCertificateSecretNameSuffix)`

SaaldjorMike: Also, `SAML_IDP_CERTIFICATE` is automatically appended to the env vars when saml is used.

jswoods: Ah got it. Will fix

Issue #116, opened 2020-07-10
Service account not updated when humioServiceAccountAnnotations are added


jswoods:
Once a HumioCluster is created along with a service account, that service account cannot be updated by adding `humioServiceAccountAnnotations`. Additionally, deleting the service account manually is not enough as the secret also needs to be updated.

Should we have an ensureMismatchedServiceAccountIsDeleted in the case where a user adds `humioServiceAccountAnnotations` to the HumioCluster spec after the initial creation? If we do this, we'd also need to re-create the secret.

jswoods:
Note that the secret associated with a service account is deleted by a cleanup job if it doesn't have a service account owner.

PR #120, opened 2020-07-21
Use dynamic suffix for service account secrets and allow updates of service account annotations

jswoods:
This pr does two things:

1) Move to dynamic suffixes for the secrets created for the service accounts. The reasoning here is if these secrets change, the pods will still attempt to use the old secrets. We rely on the dynamic suffix to trigger pod deletions to pick up the new secret name. This is not directly related to the other change in this PR, but something I noticed during my testing and something we should change now as it will be more painful later on. If you look at secrets that are type of `kubernetes.io/service-account-token` they are always named with a hash suffix. 

2) Allow updates of the service account annotations. 

Another option I explored for this was instead of doing an update on the service account resource when we change annotations, we move the service account to a dynamic/hash suffix and then do a delete/recreate similar to how we handle pods. In this case, the service account is recreated and so is the secret associated with it, however this would also trigger a recreation the pods. In the case of adding the annotation `eks.amazonaws.com/role-arn` (which is what we use this for), we do not require that we recreate the service account or secret, hence why we can just do the update. But we still need to manually delete the pods. Perhaps there may be other cases where we do need to recreate the service account and/or secret when changing a service account. In this case, because of the work of #1 in the PR, we have the option of recreating the service account and then the secret and pods will also be recreated.

So my questions are: should we move the service account to use a dynamic suffix like the secrets? Or should we leave the service account statically named but perhaps we do something to the pod to trigger recreation when the service account is changed, like delete/recreate the service account secret? Or do we leave the update logic and rely on manually deleting the pods?


jswoods:
Goal if this is to fix https://github.com/humio/humio-operator/issues/116.

jswoods:
Let's continue with this and once we have a fix for https://github.com/humio/humio-operator/issues/122 (which should be soon, as I'm working on another PR for restart functionality) then I will make the updates so that we roll the humio pods after changing the service account annotations (added a TODO for this).

jswoods:
I have rebased and refactored a bit after pulling in the restart functionality. This should be ready to go now.

deploy/crds/core.humio.com_humioclusters_crd.yaml
```diff
@@ -2,12 +2,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: humioclusters.core.humio.com
-  labels:
-    app: 'humio-operator'
-    app.kubernetes.io/name: 'humio-operator'
-    app.kubernetes.io/instance: 'humio-operator'
-    app.kubernetes.io/managed-by: 'Helm'
-    helm.sh/chart: 'humio-operator-0.0.6'
```
SaaldjorMike, line 10: Hmm. Those labels should get added automatically when running "make crds". Did you generate the CRD's using that make target or did you run `operator-sdk generate crds` directly?

pkg/controller/humiocluster/humiocluster_controller.go
```diff
@@ -1381,8 +1470,13 @@ func (r *ReconcileHumioCluster) ensurePodsExist(ctx context.Context, hc *corev1a
 		return reconcile.Result{}, err
 	}
 
+	attachments, err := r.newPodAttachments(ctx, hc, foundPodList)
+	if err != nil {
+		r.logger.Errorf("failed to get pod attachments: %s", err)
```
SaaldjorMike, line 1475: @jswoods I think we might be missing a return statement here. I assume we'd want to return the error?

deploy/crds/core.humio.com_humioclusters_crd.yaml
jswoods: fixed

pkg/controller/humiocluster/humiocluster_controller.go
jswoods: fixed

pkg/controller/humiocluster/humiocluster_controller.go
```diff
@@ -1350,7 +1434,12 @@ func (r *ReconcileHumioCluster) ensurePodsBootstrapped(ctx context.Context, hc *
 
 	r.logger.Debugf("pod ready count is %d, while desired node count is %d", podsReadyCount, hc.Spec.NodeCount)
 	if podsReadyCount < hc.Spec.NodeCount {
-		err = r.createPod(ctx, hc, foundPodList)
+		attachments, err := r.newPodAttachments(ctx, hc, foundPodList)
+		if err != nil {
+			r.logger.Errorf("failed to get pod attachments: %s", err)
+			return reconcile.Result{}, err
```
SaaldjorMike, line 1440: @jswoods Given that the other `ensurePodsExist` have `return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, err`, perhaps we want to update this one to do the same?

Issue #122, opened 2020-07-24
Add option to restart humio pods


jswoods:
There is at least one case where we need to roll the cluster following a change. See https://github.com/humio/humio-operator/issues/116. If we update the annotations on the service account, then we need to recreate all the humio pods that use the service account.

Ideally we'd have a way to roll the cluster as well as perform non-rolling restarts in the case where those are required by updating the humio version.

PR #123, opened 2020-07-24
Add humio pod restart and upgrade functionality

jswoods:
This adds support for two ways of restarting pods: Restarting and Upgrading. It also introduces the idea of a "pod revision", which allows us to trigger restarts of Humio pods without changing the HumioCluster spec.

When the HumioCluster Spec is changed such that the Humio image is updated, then we change to a cluster state of Upgrading and take down all pods until there are none running, then start them all back up mostly at the same time.

When the HumioCluster Spec is changed in any other way other than updating the image, we change to a cluster state of Restarting and take down one pod at a time, wait for it to become Ready and then move on to the next pod (similar to basic deployments).

If we want to trigger a restart of Humio without changing the spec, we can increment the revision in the annotation `humio.com/pod-revision`. We can control whether the pods are restarting in a rolling or non-rolling (recreate) fashion by setting the annotation `humio.com/pod-restart-policy` to either `restart` or `recreate`.

I'm thinking we will want at least two e2e tests for this, one for rolling and another for recreating, but I am holding off on that for now as I know you are refactoring those a bit.

### DIFF
--- a/pkg/controller/humiocluster/humiocluster_controller.go
+++ b/pkg/controller/humiocluster/humiocluster_controller.go
@@ -139,6 +139,7 @@ func (r *ReconcileHumioCluster) Reconcile(request reconcile.Request) (reconcile.
 			r.logger.Infof("unable to set cluster state: %s", err)
 			return reconcile.Result{}, err
 		}
+		r.incrementHumioClusterPodRevision(context.TODO(), hc, PodRestartPolicyRolling)
 	}
 
 	// Ensure service exists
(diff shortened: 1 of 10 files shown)


jswoods:
Fixes https://github.com/humio/humio-operator/issues/122

pkg/controller/humiocluster/humiocluster_controller.go
```diff
@@ -835,79 +838,100 @@ func (r *ReconcileHumioCluster) ensureServiceExists(ctx context.Context, hc *cor
 }
 
 // ensureMismatchedPodsAreDeleted is used to delete pods which container spec does not match that which is desired.
-// If a pod is deleted, this will requeue immediately and rely on the next reconciliation to delete the next pod.
-// The method only returns an empty result and no error if all pods are running the desired version,
-// and no pod is currently being deleted.
+// The behavior of this depends on what, if anything, was changed in the pod. If there are changes that fall under a
+// rolling update, then the pod restart policy is set to PodRestartPolicyRolling and the reconciliation will continue if
+// there are any pods not in a ready state. This is so replacement pods may be created.
+// If there are changes that fall under a recreate update, the the pod restart policy is set to PodRestartPolicyRecreate
```
SaaldjorMike, line 844:
```suggestion
// If there are changes that fall under a recreate update, the pod restart policy is set to PodRestartPolicyRecreate
```

SaaldjorMike:
I don't think the E2E code refactoring blocks us from just adding tests similar to what we've done so far. I think we should just go ahead and get the additional tests added.

Issue #113, opened 2020-07-10
Expose type for service resource


jswoods:
Expose the service type for the service resource that is created by the operator. Currently this is hard coded to `ClusterIP`.

PR #130, opened 2020-08-04
Expose options for humio service type and ports

jswoods:
Fixes https://github.com/humio/humio-operator/issues/113 and also exposes overriding ports for the service.


SaaldjorMike:
@jswoods Overall lgtm, but we probably want to fix the e2e test which for some reason fails.

Issue #152, opened 2020-09-01
Being able to configure path or documenting nginx rewrite rules for using alternate URL prefixes


andrem123456:
I would like to access my humio URL using a different URL prefix. Changing the `path` in the humio operator didn't work but this is probably possible using nginx rewrite rules.

> `nginx.ingress.kubernetes.io/rewrite-target`
> annotation to achieve the /logs path, documentation for that annotation is available here: https://kubernetes.github.io/ingress-nginx/examples/rewrite/
> 

Issue to track either documenting rewrite rule or adding support for the `path` in the Humio Operator. Preferabbly being able to do via the `path` in Humio operator would be ideal

nevosial:
Please note even if we use the rewrite rules; we cannot achieve the desired result.

As we try to hit the <host-name>/logs  --> `The repository 'logs' might actually exist but you lack permission to access it. If the URL seems correct, try logging in as another user or get a repository admin to grant you access.` 
Clicking on the `Go to front page` button redirects to the <host-name> i.e Humio UI.

PR #154, opened 2020-09-04
Add support for running Humio on non-root paths

jswoods:
Fixes https://github.com/humio/humio-operator/issues/152


schofield (approved):

SaaldjorMike (approved):
Maybe we should also add an example in https://github.com/humio/humio-operator/tree/master/examples ?

PR #156, opened 2020-09-04
Reduce labels and selector matchLabels that the operator deployment uses

jswoods:
When using spinnaker, it sets the label `"app.kubernetes.io/managed-by"`, which then causes the deployment to break due to:

```
Deploy failed: The Deployment "humio-operator" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"app":"humio-operator", "app.kubernetes.io/instance":"humio-operator", "app.kubernetes.io/managed-by":"spinnaker", "app.kubernetes.io/name":"humio-operator"}: selector does not match template labels
```

Once the managed-by label is removed, it works as the selector is no longer using the label, thus it no longer needs to match. I don't think any of these are necessary on the operator deployment? The downside of changing these labels is the deployment will need to be deleted prior to updating to the version of the operator that contains this change.

### DIFF
--- a/charts/humio-operator/templates/operator-deployment.yaml
+++ b/charts/humio-operator/templates/operator-deployment.yaml
@@ -20,9 +20,9 @@ spec:
     type: Recreate
   selector:
     matchLabels:
+      app: '{{ .Chart.Name }}'
       app.kubernetes.io/name: '{{ .Chart.Name }}'
       app.kubernetes.io/instance: '{{ .Release.Name }}'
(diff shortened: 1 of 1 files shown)


jswoods:
This is not changing the labels of the operator, but the pods created by the operator. I'm not sure if they fall under the same requirements?

SaaldjorMike:
I was under the impression that all pods needs these annotations, including the operator itself and whatever resources it creates. But, I guess if this helps out for now, we could go ahead with this and revisit this later on when we have another linter run.

SaaldjorMike (approved):

PR #159, opened 2020-09-19
Add chart repo install instructions

Crevil:
When trying to setup the operator I couldn't get the installation instructions in this readme to work as the chart repo was unknown.

This adds the `helm repo add` command to the installation instructions, which fixed it for me.

It is done the same way as described in https://github.com/humio/humio-operator/blob/master/docs/README.md#optional-prepare-an-installation-of-kafka-and-zookeeper


SaaldjorMike:
Thanks

PR #161, opened 2020-10-01
Upgrade to operator-sdk 1.0.1

SaaldjorMike:


charts/humio-operator/templates/crds.yaml
```diff
@@ -1,93 +1,12 @@
...
-                type: string
-            type: object
-        type: object
-    served: true
-    storage: true
-    subresources:
-      status: {}
----
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
```
jswoods, line 4: Do you know why we moved from `v1` to `v1beta1` here?

jswoods (approved):

SaaldjorMike:
All I know is that it is due to the new scaffolding logic. Perhaps they are
using the old beta API to have better backwards compatibility? Not entirely
sure.

PR #169, opened 2020-10-02
Define only one logger on Reconciler structs

### LINKED ISSUE #168: Cleanup use of loggers
Right now we have a mix of different loggers in the code base, and we need to clean this up.

For instance, all our reconciler structs have two loggers:

```go
// HumioClusterReconciler reconciles a HumioCluster object
type HumioClusterReconciler struct {
	client.Client
	Log         logr.Logger // TODO: Migrate to *zap.SugaredLogger
	logger      *zap.SugaredLogger
	Scheme      *runtime.Scheme
	HumioClient humio.Client
}
```

We should be able to stick to one logger.

SaaldjorMike:
Fixes #168 

@jswoods Logs are still identical the old logs, or at least very close to, but at least this way we only define one logger on the reconciler structs.

It is possible we could replace the `logr.Logger` with the `zap.SugaredLogger` on the reconciler's, but the default scaffolding assumes we use `logr.Logger`. Let me know what you think we should do. I'm thinking we should just go with this for now and perhaps refactor this to the other logger later if needed. Personally, I really like the way we log errors with the approach shown in this PR, and I think we log out errors many more places than we do info logs, so perhaps we can live without the old `Infof()` method and such.

### DIFF
--- a/controllers/humiocluster_annotations.go
+++ b/controllers/humiocluster_annotations.go
@@ -40,7 +40,7 @@ func (r *HumioClusterReconciler) incrementHumioClusterPodRevision(ctx context.Co
 		return -1, err
 	}
 	newRevision++
-	r.logger.Infof("setting cluster pod revision to %d", newRevision)
+	r.Log.Info(fmt.Sprintf("setting cluster pod revision to %d", newRevision))
 	hc.Annotations[podRevisionAnnotation] = strconv.Itoa(newRevision)
 
 	r.setRestartPolicy(hc, restartPolicy)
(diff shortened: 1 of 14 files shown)


controllers/humioresources_controller_test.go
```diff
@@ -337,6 +337,7 @@ var _ = Describe("Humio Resources Controllers", func() {
 				err := k8sClient.Get(context.Background(), key, fetched)
 				return errors.IsNotFound(err)
 			}, testTimeout, testInterval).Should(BeTrue())
+			//Expect(err).ToNot(Succeed()) // This will fail the test
```
jswoods, line 340: Should this be included in this PR?

jswoods (approved):
This seems good to me. I don't see any reason we need to use the SugaredLogger right now. And missing the Infof doesn't seem like that big of a deal.

controllers/humioresources_controller_test.go
SaaldjorMike: Good catch, no. I'll remove it. Thanks!

PR #177, opened 2020-10-08
Wait a bit before requeue after creating a PVC.

SaaldjorMike:


jswoods (approved):
Did you confirm this had any effect on the test? I'd be surprised if this fixed it, since the RequeueAfter doesn't force any type of pause (it is misleading). I think we likely want something like waitForNewPod() but for the pvc, e.g. waitForNewPVC(). Though this is definitely worth a try.

SaaldjorMike:
> Did you confirm this had any effect on the test?

I only reran the tests on github actions a bunch of times. I've not been able to reproduce it locally so far.

PR #183, opened 2020-10-09
sidecar: Drop cluster name and namespace from the hostname

SaaldjorMike:
This way we do not rely on cluster DNS. We can do this
because `/etc/hosts` for the sidecar already contains the
information needed to perform the lookup

### DIFF
--- a/images/helper/main.go
+++ b/images/helper/main.go
@@ -137,6 +137,33 @@ func createAndGetAdminAccountUserID(client *humio.Client) (string, error) {
 	return "", fmt.Errorf("could not obtain user ID")
 }
 
+// validateAdminSecretContent grabs the current token stored in kubernetes and returns nil if it is valid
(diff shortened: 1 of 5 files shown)


images/helper/main.go
```diff
@@ -137,6 +137,33 @@ func createAndGetAdminAccountUserID(client *humio.Client) (string, error) {
...
+
+		_, err = humioClient.Clusters().Get()
+		if err != nil {
+			return err
+		}
+
+		// We could successfully get information about the cluster, so the token must be valid
+		return nil
+	}
+	return fmt.Errorf("unable to validate if kubernetes secret %s holds a valid humio api token", adminSecretName)
```
jswoods, line 164: Not a big deal, but this error is never printed

jswoods (approved):

PR #190, opened 2020-10-13
fix: helm2 compatibility

estahn:
Helm2 produces an invalid manifest:

```yaml
# Source: humio-operator/templates/operator-rbac.yaml
---apiVersion: v1
kind: ServiceAccount
metadata:
  name: 'release-name'
```


SaaldjorMike (approved):
Thanks!

PR #218, opened 2020-10-26
Fix test timeouts

jswoods:
Fixes a number of issues including the timeouts in `waitForPod()` when running tests, race condition between the pod certificate and the pod creation, and PVC support for migrating from ephemeral to PVCs.


controllers/humiocluster_tls.go
```diff
@@ -199,3 +207,63 @@ func constructNodeCertificate(hc *humiov1alpha1.HumioCluster, nodeSuffix string)
...
+
+			// Check if we should update the existing certificate
+			certForHash := constructNodeCertificate(hc, "")
+
+			// Keystores will always contain a new pointer when constructing a certificate.
+			// To work around this, we override it to nil before calculating the hash,
+			// if we do not do this, the hash will always be different.
+			certForHash.Spec.Keystores = nil
+
+			desiredCertificateHash := helpers.AsSHA256(certForHash)
```
SaaldjorMike, line 245: Perhaps we could do the same json marshal here before calculating the hash?

controllers/humiocluster_tls.go
```diff
@@ -199,3 +207,63 @@ func constructNodeCertificate(hc *humiov1alpha1.HumioCluster, nodeSuffix string)
...
+		r.Log.Info(fmt.Sprintf("validating new pod certificate was created. expected pod certificate count %d, current pod certificate count %d", expectedCertCount, existingNodeCertCount))
+		if existingNodeCertCount >= expectedCertCount {
+			return nil
+		}
+		time.Sleep(time.Second * 1)
+	}
+	return fmt.Errorf("timed out waiting to validate new pod certificate was created")
+}
+
+func (r *HumioClusterReconciler) nodeCertificateCount(ctx context.Context, hc *humiov1alpha1.HumioCluster) (int, error) {
```
SaaldjorMike, line 226: I don't think the method name is good at conveying what it is doing. To me this "count" suffix makes me think that it just returns the number of certificates we have right now, but doesn't touch the resources. This looks to not be true as it does trigger updates of certificates if their hash differs.

Perhaps it would be fine if we'd just add some godoc for it, but I think we should see if we can find a more suitable method name.

SaaldjorMike (approved):
Added a few minor comments, but LGTM. I'll leave it up to you to decide if you want to fix those before merging.

controllers/humiocluster_tls.go
jswoods: Right, this was a poor name choice.

controllers/humiocluster_tls.go
jswoods: Yes.

PR #230, opened 2020-10-29
Add test script with shared ServiceAccount and linkerd proxy injection

kaspernissen:
This PR adds a test script and sample `HumioCluster` with `linkerd` proxy injection using `podAnnotations` and a shared `ServiceAccount`. 

The script simulates the manual creation of resources when opting out of the service account per container setup and setup of linkerd in a kind cluster.


SaaldjorMike (approved):

jswoods (approved):

jswoods:
Awesome, thank you for this! We also have a helm chart test script which we plan on running via a CI job. I think we should also add this test when we do that work.

kaspernissen:
No problem. Would be awesome to have this scenario run in CI somehow 👏

PR #232, opened 2020-10-29
Bump default Humio version to be latest stable

SaaldjorMike:


ssboisen (approved):
LGTM - I'll make the necessary changes to the workflow in humio/humio

jswoods (approved):

PR #282, opened 2020-11-18
Validate API token is obtained using the expected methods

SaaldjorMike:
I did run the tests locally where I bumped the default Humio container image tag to 1.17.0, which worked just fine, but I'm guessing we want to stick with our stable releases as the default version.

Also, not sure what to do about the constants at this point. Right now the helper image only contains one package, which I can't import, so we'd need to split that up if we want to keep the constants in a separate package. If we split such things out, we may also want to refactor the other couple of things we have layout around which also has a copy in the main project as well as the helper image project.

### DIFF
--- a/controllers/humiocluster_controller_test.go
+++ b/controllers/humiocluster_controller_test.go
@@ -38,6 +38,16 @@ import (
 
 const autoCleanupAfterTestAnnotationName = "humio.com/auto-cleanup-after-test"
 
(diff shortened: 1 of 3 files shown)


jswoods:
One thing that comes to mind is that we could have some generator code where we write out the constants, or any other code that is shared between the main project and the helper to both modules without requiring imports/dependencies.

That doesn't necessarily need to be done as part of this PR.

jswoods (approved):

Issue #165, opened 2020-10-02
Leverage Humio's suggested partition layouts


SaaldjorMike:
Right now we have our init container figure out what zone the pod was deployed in, and then it pass that on to Humio using a shared volume which is read by the Humio container before starting up Humio itself.

We should be able to do something similar as I describe above, to set the `ZONE` configuration option so we can leverage the new functionality in Humio 1.15+ that can suggest partition layouts for storage and digest partitions.

With the recent merge of https://github.com/humio/humio-operator/pull/174 we now set `ZONE`, but still need to set the replication factor options to leverage suggested partition layouts.

PR #284, opened 2020-11-23
Use suggested partition layouts for Humio 1.17+

SaaldjorMike:
This also introduces a new field which makes it possible to disable init container entirely, which can be used if running in a single-AZ Kubernetes cluster.

Fixes https://github.com/humio/humio-operator/issues/165


jswoods (approved):
awesome! 🎉

PR #298, opened 2020-12-04
Only leverage Zookeeper to obtain node UUID when using ephemeral disks

SaaldjorMike:
For existing clusters using this in combination with persistent disks,
they will continue to work as they will reuse the existing node UUID
file from the persistent disk. This means that no changes are required
by the user here even if the default behaviour changes slightly.

### DIFF
--- a/controllers/humiocluster_controller_test.go
+++ b/controllers/humiocluster_controller_test.go
@@ -614,20 +614,24 @@ var _ = Describe("HumioCluster Controller", func() {
 	})
 
 	Context("Humio Cluster Container Arguments", func() {
-		It("Should correctly configure container arguments", func() {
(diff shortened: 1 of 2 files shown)


ssboisen (approved):
LGTM - good way to solve the problem!

jswoods (approved):
Looks good. Do you think we should add a couple expect statements for this to this test https://github.com/humio/humio-operator/blob/master/controllers/humiocluster_controller_test.go#L616-L654 and rename it to something like "Correctly configure container arguments and ephemeral disks env var"?

SaaldjorMike:
> Do you think we should add a couple expect statements for this to this test

Sounds like a good idea. Will update this PR with that.

Issue #238, opened 2020-10-30
HumioCluster should have license attribute


estahn:
I would like to be able to define the license via `HumioCluster`, e.g.

```yaml
apiVersion: core.humio.com/v1alpha1
kind: HumioCluster
metadata:
  name: humio
  namespace: logging
spec:
  image: "humio/humio-core:1.16.0"
  license: "abce..."
```
or/and
```yaml
apiVersion: core.humio.com/v1alpha1
kind: HumioCluster
metadata:
  name: humio
  namespace: logging
spec:
  image: "humio/humio-core:1.16.0"
  license:
    secretKeyRef:
      name: humio-license
      key: data
```

PR #315, opened 2020-12-18
Manage licenses through the humiocluster resource

jswoods:
Fixes https://github.com/humio/humio-operator/issues/238


jswoods:
Will need to add docs for:
```
kubectl create secret generic example-humiocluster-license --from-literal=data=<license>
```

And then adding to the spec:
```
+  license:
+    secretKeyRef:
+      name: example-humiocluster-license
+      key: data
```

controllers/humiocluster_controller.go
```diff
@@ -1168,6 +1173,74 @@ func (r *HumioClusterReconciler) ensurePvcLabels(ctx context.Context, hc *humiov
...
+		return reconcile.Result{}, nil
+	}
+
+	licenseSecret, err := kubernetes.GetSecret(ctx, r, licenseSecretKeySelector.Name, hc.Namespace)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.Log.Error(err, fmt.Sprintf("license was requested but no secret exists by name %s in namespace %s", licenseSecretKeySelector.Name, hc.Namespace))
+			return reconcile.Result{}, err
+		}
+	}
```
SaaldjorMike, line 1203: Looks like we right now ignore any errors that are not true when calling `errors.IsNotFound()`. Is that intentional?

controllers/humiocluster_controller.go
```diff
@@ -1168,6 +1173,74 @@ func (r *HumioClusterReconciler) ensurePvcLabels(ctx context.Context, hc *humiov
...
+		}
+
+		r.Log.Info(fmt.Sprintf("successfully installed license: type: %s, issued: %s, expires: %s",
+			desiredLicense.LicenseType(), desiredLicense.IssuedAt(), desiredLicense.ExpiresAt()))
+
+		// refresh the existing license for the status update
+		existingLicense, err = r.HumioClient.GetLicense()
+		if err != nil {
+			r.Log.Error(err, "failed to get updated license: %v", err)
+		}
```
SaaldjorMike, line 1237: It seems slightly weird to me that we have these lines here, given that they are not important for ensuring the license is installed to the Humio cluster. I feel like they should be put closer to the code that actually updates the license status. Right now it is not clear when reading these few lines, what other parts of the code depends on it.

controllers/humiocluster_controller_test.go
```diff
@@ -2169,6 +2179,66 @@ var _ = Describe("HumioCluster Controller", func() {
 			}, testTimeout, testInterval).Should(BeEquivalentTo(120))
 		})
 	})
+
+	Context("Humio Cluster install license", func() {
+		It("Should succesfully install a license", func() {
```
SaaldjorMike, line 2184: Perhaps we should also include a test case where it tries installing an invalid license? Perhaps the HumioCluster.Status.State should even result in `ConfigError` if the referenced k8s secret does not exist, or if it does not contain the expected content?

controllers/humiocluster_status.go
```diff
@@ -56,6 +56,15 @@ func (r *HumioClusterReconciler) setVersion(ctx context.Context, version string,
 	}
 }
 
+func (r *HumioClusterReconciler) setLicense(ctx context.Context, licenseStatus humiov1alpha1.HumioLicenseStatus, hc *humiov1alpha1.HumioCluster) {
```
SaaldjorMike, line 59: I'm starting to think we should rename the functions inside this file. Right now it is not that clear what `setLicense` actually means.

To me, the name indicates that it sets the license in the Humio cluster, similar to if the function was called `activateLicense` or perhaps `installLicense`.

This is probably a more general thing about all the functions we have that sets/updates the `HumioCluster.Status` fields, so I'm also good with keeping it the way it is for now, but I do think we should consider doing something to make it more clear when reading the code that these functions are "only" updating the object status.

controllers/humiocluster_controller_test.go
```diff
@@ -106,6 +107,15 @@ var _ = Describe("HumioCluster Controller", func() {
 					}
 
 					_ = k8sClient.Delete(context.Background(), &cluster)
+
+					if cluster.Spec.License.SecretKeyRef != nil {
+						k8sClient.Delete(context.Background(), &corev1.Secret{
```
SaaldjorMike, line 112: I'm considering if we should introduce a linting rule such that we enforce return values to be assigned to a variable. The main goal for that is to make sure we don't accidentally skip handling an error if we should have.

That would mean this line instead would require this change, since we don't really care about the output of deleting the object in this specific line:
```suggestion
						_ = k8sClient.Delete(context.Background(), &corev1.Secret{
```
I'm not sure about the details about how to actually enforce it, but what do you think about introducing such a rule?

controllers/humiocluster_controller.go
jswoods: I agree it would be nice to clean up the main Reconcile function. There are a number of status update helper functions that are not as easily correlated to an ensure method like license is. But perhaps we should do some refactoring in this area.

controllers/humiocluster_status.go
jswoods: Yes I agree, and a refactor in general should be done to make a better pattern for the status updates.

controllers/humiocluster_controller_test.go
jswoods: Linting sounds good to me. Perhaps another CI job?

SaaldjorMike (approved):

PR #332, opened 2021-02-05
Installs ubi security updates during build

schofield:
Because we use latest I'm not sure often thing will be upgraded but this was suggested a best practice.


jswoods (approved):

schofield:
Read the docs.

PR #338, opened 2021-02-27
Add support for actions and alerts

jswoods:


main.go
```diff
@@ -38,6 +38,7 @@ import (
 	"github.com/humio/humio-operator/pkg/helpers"
 	"github.com/humio/humio-operator/pkg/humio"
 
+	corev1alpha1 "github.com/humio/humio-operator/api/v1alpha1"
```
SaaldjorMike, line 41:
```suggestion
```

main.go
```diff
@@ -51,6 +52,7 @@ func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
 	utilruntime.Must(humiov1alpha1.AddToScheme(scheme))
+	utilruntime.Must(corev1alpha1.AddToScheme(scheme))
```
SaaldjorMike, line 55:
```suggestion
```

main.go
SaaldjorMike: We end up registering the same thing twice, that is probably not the intent here.

SaaldjorMike (approved):
LGTM.

PR #399, opened 2021-06-29
Reuse HTTP connections

SaaldjorMike:
This PR ensures that the operator will reuse the HTTP connections when communicating with the humio cluster.
This also updates the auth sidecar to reuse the connection it needs to periodically validate the API token works.

I ran a couple of tests with the new helper changes, to make sure they work, then I reverted back to the previous default helper image which is what the current state of this PR is at. My plan is to cut a new helper image release after merging this PR, then bump the default helper image to this new version.


pkg/humio/client.go
```diff
@@ -106,37 +107,53 @@ type LicenseClient interface {
...
+		c = humioapi.NewClient(*config)
+	} else {
+		existingConfig := c.Config()
+		equal := existingConfig.Token == config.Token &&
+			existingConfig.Insecure == config.Insecure &&
+			existingConfig.CACertificatePEM == config.CACertificatePEM &&
+			existingConfig.ProxyOrganization == config.ProxyOrganization &&
+			existingConfig.Address.String() == config.Address.String()
+		if !equal {
+			//h.logger.Info("recreating Humio API client")
```
jswoods, line 151: Do you mean to have this commented out?

jswoods (approved):

pkg/humio/client.go
SaaldjorMike: Ah, no. I commented that out with the intention to delete it later, but looks like I forgot to do that. I've force pushed a commit that gets rid of this line. Thanks!

PR #421, opened 2021-08-13
Update Humio client config before validating/installing license

SaaldjorMike:
Without this, the Humio client config might point to a different
HumioCluster/HumioExternalCluster resource, from the previous
reconciliation.


ssboisen (approved):
LGTM - This will make it easier when running test deployments!

jswoods (approved):

PR #425, opened 2021-09-02
Upgrade kind to 1.21.2

jswoods:


SaaldjorMike:
We have two ways of running the e2e test suite. One that spins up `kind` with this script you update here, but we also have github workflows where we use this step with a dedicated action https://github.com/humio/humio-operator/blob/master/.github/workflows/e2e.yaml#L9-L12 to spin up `kind`. Looks like they didn't match before. Did you intentionally only bump one of them? I think I might've bumped on of them previously and forgot there's two places where the `kind` versions are set.

SaaldjorMike (approved):

jswoods:
Ah - that makes sense. It was failing for me locally so I had to do the upgrade here. The 1.17 version no longer works since this update https://github.com/humio/humio-operator/commit/bfe165b1e88a89821c2d79b986f1b997a4579c5c.

The build matrix is a good idea, but it will make the tests that much longer. Should we do that as part of this PR or a separate one? Perhaps I should also bump the actions test to use kind 1.21.2 as well.

SaaldjorMike:
> The build matrix is a good idea, but it will make the tests that much longer. Should we do that as part of this PR or a separate one? If we do it as a separate task, then perhaps for this one I should also bump the actions test to use kind 1.21.2 as well.

The matrix thing is probably worth doing separately. And sure, lets go with 1.21.2 both places for now.

PR #430, opened 2021-09-13
Create a single base logger and attach version information to log entries

SaaldjorMike:
This together with function names and caller line numbers makes it easy
to know exactly where things are originating from.


ssboisen (approved):
great change!

jswoods (approved):

PR #431, opened 2021-09-13
Fix bug where pod name is not logged during pod termination

jswoods:


SaaldjorMike (approved):

ssboisen (approved):
LGTM - Should improve the deployment flow a great deal, especially on big clusters.

PR #437, opened 2021-09-21
Adjust how license code connects to the humio cluster and add additonal license logging

jswoods:


schofield (approved):

controllers/humiocluster_controller.go
```diff
@@ -1515,17 +1515,32 @@ func (r *HumioClusterReconciler) ensureLicenseIsValid(ctx context.Context, hc *h
...
+			return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 10}, nil
+		}
+		if err != nil {
+			r.Log.Error(err, "unable to obtain CA certificate")
+			return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 10}, err
+		}
+		humioAPIConfig.CACertificatePEM = string(existingCABundle.Data["ca.crt"])
 	}
-	r.HumioClient.SetHumioClientConfig(cluster.Config(), req)
+	r.HumioClient.SetHumioClientConfig(humioAPIConfig, req)
```
ssboisen, line 1535: I think we should configure the HumioClient in the topmost part of the Reconcile method, right after we fetch the HC configuration.

jswoods: I don't think we can do that since we must wait for the certs, but we can probably refactor this into a method.

PR #438, opened 2021-09-22
Leave Humio's tmp dir on the same mount as humio-data

SaaldjorMike:


jswoods (approved):

ssboisen (approved):
LGTM though I think we should consider changing how we do humio version comparison to make it easier to read, I would prefer if it said something like

```golang
if humioVersion.Before(HumioVersionWithNewTmpDir) {
...
}
```
to me that's much easier to reach then having to look at the atleast assignment and then negate that in my head when trying to understand what the condition needs to be for this to be true

jswoods (approved):

SaaldjorMike:
> LGTM though I think we should consider changing how we do humio version comparison to make it easier to read, I would prefer if it said something like
> 
> ```go
> if humioVersion.Before(HumioVersionWithNewTmpDir) {
> ...
> }
> ```
> 
> to me that's much easier to reach then having to look at the `AtLeast` assignment and then negate that in my head when trying to understand what the condition needs to be for this to be true

That would certainly be easier to read. For now, I'd like to stick with this approach and then we can improve it a little later given how few places we do this.

PR #445, opened 2021-09-28
Add *Source variants to HumioAction resource

### LINKED ISSUE #351: Ability to add config params via secrets in humio actions
In the humioAction resource, particularly the slack notifier,
when we use multi-channel slack action; it expects a `Slack OAuth Token`  ; 
Can this token be mounted from a secret on the k8s cluster instead of specifying it in the humioAction manifest itself ?

Is something similar possible for the single channel action as well…

Basically ability to mount the `slackProperties:` and `slackPostMessageProperties:`  from a configMap or a secret on the k8s cluster.

Once this feature is enabled we can happily version control our Humioaction manifest without worrying about exposing secrets.

Crevil:
Currently it is not possible to reference kubernetes Secret resources for secret
fields like ingest tokens and API keys.

This change adds `*Source` variants to these fields where a secret can be
referenced.

This fixes #351


jswoods:
Hi @Crevil - thank you for this! As you can see, we have a recent change in this project which causes our e2e tests for fail for forks. This needs to be fixed, but I'd rather not wait for that to be fixed before we can merge this, so I pushed a separate PR for you fork just so we can run the tests and then decide if we can go ahead and merge this PR.

Crevil:
Thanks. 🙌 I'm unsure whether I need to do anything here?

SaaldjorMike:
@Crevil Yeah, we don't expect you to do more at this stage. We're currently struggling with some flaky tests which we'd like to fix first. It is not something that was introduced by your work, we just want to get it fixed before we run a new round of tests based on your work. Then we just need this change rebased on top of master once the flakiness has been resolved.

controllers/humioaction_controller_test.go
```diff
@@ -0,0 +1,169 @@
...
+var _ = Describe("Humio Action Controller", func() {
+	BeforeEach(func() {
+		clusterKey := types.NamespacedName{
+			Name:      "humiocluster-shared",
+			Namespace: "default",
+		}
+		cluster := constructBasicSingleNodeHumioCluster(clusterKey, true)
+		ctx := context.Background()
+		createAndBootstrapCluster(ctx, cluster, true)
+	})
```
SaaldjorMike, line 40: Running `make test` where we do things in parallel causes this to conflict with https://github.com/humio/humio-operator/blob/master/controllers/humioresources_controller_test.go

I like the idea of moving things into separate files for testing the different controllers, but on the other hand we also need to consider whether we really _need_ a fresh/new Humio cluster instance since they take time to spin up (and take down), usually we something like ~60 seconds for single node clusters. So to keep things consistent for now, we probably want to move the contents of the separate`It()` cases in this file, into the single (and very long) `It()` inside https://github.com/humio/humio-operator/blob/master/controllers/humioresources_controller_test.go
Then later on, we may decide on breaking them up a little more. The test suite with `kind` is already up to something like ~1 hour of runtime, so I'd rather not see more Humio clusters we spin up to test this sort of logic until we've solved the issues that are related to running test suites in parallel. Once we can run test-cases in parallel, then I suspect we will rip out the specific controller tests to each their own `<crdName>_controller_test.go` files to essentially follow what you're trying to do here.

Crevil: Got it 👍 

Should I just add these tests to after the existing action tests in that file?

SaaldjorMike: Yeah, let's do that.

I'm hoping that we will soon start running tests in parallel so we can break out these cases in their separate `.go` files, but we're not quite there at this point.

Thanks!

SaaldjorMike: Hmm. Looks like a test didn't pass, but it also doesn't seem like a test related to this PR. Perhaps we could rebase this once more? I just merged a fix to master which should also make the e2e's slightly faster (but still takes~40-45 mins compared to perhaps ~60-70 mins before).

Do note that there is a slight change to the tests where the namespace we create resources like `HumioAlert` in must be done in the correct namespace. You can see the test cases that's on master to see what should be used for namespace instead of setting `Namespace: "default"`. After rebasing "fail-fast" will be disabled so we can see if this is only a single job failing or if all of them fails. Before rebasing it cancels all the other jobs once the first one has failed. _Also_ we've now included k8s 1.22 so we run tests for k8s 1.19 through 1.22 when testing so we now have a 4 full test runs to check out the results for.

Hopefully we will soon have a much better process of handling external contributions like these. Ideally we would be able to also easily kick off a round of jobs using the github secrets in _our repo_. That way it will be easier to run tests with a valid license inside github, perhaps using something like https://github.com/imjohnbo/ok-to-test or similar.

PR #449, opened 2021-09-29
Add org.opencontainers.image LABELS to docker build steps

slimslenderslacks:
cc @schofield 
Grant, based on our conversation last week, I just wanted to summarize the additional metadata that we are scanning for.  Two of the labels are part of the spec `org.opencontainers.image.revision` and `org.opencontainers.image.source`.  The third one (`com.atomist.containers.image.dockerfile`) distinguishes different Dockerfiles within one repository, and gives us the ability to do better forensics on how vulnerabilities map to different layers.

I tested this out and made sure that we're scanning and can set vulnerability baselines for our CheckRuns.

I also tested this pinning feature and unfortunately, we don't yet support pinning for public images in `registry.access.redhat.com`.  The vulnerability scanning is fine because we pull the image.  But the pinning feature is constantly scanning for moving tags (like when the `latest` tag moves on the ubi8/ubi8-minimal image.  It looks to me like they don't support the docker v2 api for pulling manifests, and even their v1 api support was missing the digest information.  We'll support it but unfortunately, you wouldn't see any automatic pinning pull requests on this repo if you turned us on here.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -22,7 +22,18 @@ jobs:
         echo "RELEASE_COMMIT=$(git rev-parse --verify HEAD)" >> $GITHUB_ENV
         echo "RELEASE_DATE=$(date --iso-8601=seconds)" >> $GITHUB_ENV
     - name: docker build
-      run: make docker-build-operator IMG=humio/humio-operator:${{ env.RELEASE_VERSION }} IMG_BUILD_ARGS="--label version=${{ env.RELEASE_VERSION }} --label release=${{ github.run_id }} --build-arg RELEASE_VERSION=${{ env.RELEASE_VERSION }} --build-arg RELEASE_COMMIT=${{ env.RELEASE_COMMIT }} --build-arg RELEASE_DATE=${{ env.RELEASE_DATE }}"
+      run: make docker-build-operator
+      env:
+        IMG: humio/humio-operator:${{ env.RELEASE_VERSION }}
+        IMG_BUILD_ARGS: >-
+          --label version=${{ env.RELEASE_VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/master.yaml
```diff
@@ -22,7 +22,18 @@ jobs:
         echo "RELEASE_COMMIT=$(git rev-parse --verify HEAD)" >> $GITHUB_ENV
         echo "RELEASE_DATE=$(date --iso-8601=seconds)" >> $GITHUB_ENV
     - name: docker build
-      run: make docker-build-operator IMG=humio/humio-operator:${{ env.RELEASE_VERSION }} IMG_BUILD_ARGS="--label version=${{ env.RELEASE_VERSION }} --label release=${{ github.run_id }} --build-arg RELEASE_VERSION=${{ env.RELEASE_VERSION }} --build-arg RELEASE_COMMIT=${{ env.RELEASE_COMMIT }} --build-arg RELEASE_DATE=${{ env.RELEASE_DATE }}"
+      run: make docker-build-operator
```
slimslenderslacks, line 25: I broke out these `env`s from the `run` because I wanted to understand what was already there.  But the only change is the addition of 3 new labels

SaaldjorMike:
@slimslenderslacks Before we merge this I have a quick question. Is it intentional that you only added these labels to the master builds of our docker images? We build our releases using other workflows, namely https://github.com/humio/humio-operator/blob/master/.github/workflows/release-container-helperimage.yaml#L25 and https://github.com/humio/humio-operator/blob/master/.github/workflows/release-container-image.yaml#L25

If it's intentional, just let us know and we can merge it as-is. If not intentional, maybe you could update this PR so it includes similar changes to those other workflows?

PR #453, opened 2021-10-01
Fix controller logic to retrun proper errors and add hack to fix race in tests

jswoods:


controllers/humiocluster_controller_test.go
```diff
@@ -911,6 +918,15 @@ var _ = Describe("HumioCluster Controller", func() {
...
 			// TODO: Right now the service is not updated properly, so we delete it ourselves to make the operator recreate the service
+			// Wait for the new HumioCluster to finish any existing reconcile loop by waiting for the
+			// status.observedGeneration to equal at least that of the current resource version. This will avoid race
+			// conditions where the HumioCluster is updated and service is deleted mid-way through a reconcile.
+			resourceVersion, _ = strconv.Atoi(updatedHumioCluster.ResourceVersion)
+			Eventually(func() bool {
+				k8sClient.Get(ctx, key, &updatedHumioCluster)
+				observedGeneration, _ := strconv.Atoi(updatedHumioCluster.Status.ObservedGeneration)
+				return observedGeneration > resourceVersion
+			}, testTimeout, testInterval).Should(BeTrue())
```
SaaldjorMike, line 929: Perhaps we should just `return observedGeneration` then use `BeNumerically()` to compare the values? Then we'd be able to see an error stating what exact values `resourceVersion` and `observedGeneration` had when the test fails instead of just the fact that it isn't true. See more here: https://onsi.github.io/gomega/#benumericallycomparator-string-compareto-interface

jswoods: Ah I was looking for that type of helper but missed that one. Thanks!

SaaldjorMike (approved):

PR #454, opened 2021-10-04
Bump humio/cli dependency and require humio 1.28.0+

SaaldjorMike:
Note: This change bumps the minimum supported humio version from 1.26.0 to 1.28.0.


jswoods (approved):
Lgtm. The only thing is since we are bumping the default humio versions as well would you mind updating the PR description and/or make sure it's noted in the commit message. I just don't want to lose track of that when doing the release notes.

PR #456, opened 2021-10-05
Use is-node-up endpoint as container probes

SaaldjorMike:


jswoods (approved):

ssboisen (approved):
Great that we get a more appropriate endpoint!

PR #499, opened 2021-11-19
Bump github.com/humio/cli dependency to fix deprecation

SaaldjorMike:


jswoods (approved):

SaaldjorMike:
@jswoods After rebasing this on top of the node pool changes I ran in to an issue with the `humiocluster-nodepool-labels` test where it seems like the `waitForNewPod()` method sometimes ends up timing out after 30 seconds, which isn't that great if the `Eventually()` step in the test case also contains a timeout of 30 seconds. For now, I've just lowered the timeout for the method which means the reconcile does actually run again and that means the label gets added back to the pod and the test now passes.

jswoods (approved):

PR #503, opened 2021-11-25
Fetch CPU core count using getconf if user didn't specify core count or resource settings

SaaldjorMike:
We do this because Kubernetes per default adds a cpu-shares flag when
running containers, and this means the amount of resources detected by
the JVM isn't what you'd expect. If no limits are set, we would expect
it to leverage the entire host and not a very tiny portion of the host.

From https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run:
```
When using Docker:

The spec.containers[].resources.requests.cpu is converted to its core
value, which is potentially fractional, and multiplied by 1024. The
greater of this number or 2 is used as the value of the --cpu-shares
flag in the docker run command.
```


jswoods (approved):

SaaldjorMike:
Needs a little more work after I rebased it, so will move to draft until I get a chance to look at it.

PR #504, opened 2021-11-25
Update auth role to only allow changes to the specific secret name that it needs to access.

SaaldjorMike:


SaaldjorMike:
Looks like this doesn't work due to https://kubernetes.io/docs/reference/access-authn-authz/rbac/
`
Note: You cannot restrict create or deletecollection requests by their resource name. For create, this limitation is because the name of the new object may not be known at authorization time. If you restrict list or watch by resourceName, clients must include a metadata.name field selector in their list or watch request that matches the specified resourceName in order to be authorized. For example, kubectl get configmaps --field-selector=metadata.name=my-configmap
`

PR #506, opened 2021-12-01
Ensure we have a storage configuration in test case

SaaldjorMike:


ssboisen (approved):
LGTM EmptyDir ftw! :)

PR #530, opened 2022-01-11
Retry failed reconciles when managing humio resources

jswoods:
Fixes an issue where if a humio resource such as a repo is created before the cluster is ready, the resource will fail to be created and not retry until the operator is restarted. This will force a retry after 15 seconds.

### DIFF
--- a/controllers/humioparser_controller.go
+++ b/controllers/humioparser_controller.go
@@ -20,14 +20,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
+	"time"
+
 	"github.com/google/go-cmp/cmp"
 	humioapi "github.com/humio/cli/api"
 	"github.com/humio/humio-operator/pkg/helpers"
 	"github.com/humio/humio-operator/pkg/kubernetes"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sort"
-	"time"
 
 	"github.com/go-logr/logr"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


controllers/humioparser_controller.go
```diff
@@ -81,7 +82,7 @@ func (r *HumioParserReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		if err != nil {
 			return reconcile.Result{}, r.logErrorAndReturn(err, "unable to set cluster state")
 		}
-		return reconcile.Result{}, err
+		return reconcile.Result{RequeueAfter: time.Second * 15}, nil
```
someword, line 85: @jswoods - does this just retry 1 time and then exit?

someword (commented):
@jswoods - Does the retry only happen 1 time?

controllers/humioparser_controller.go
jswoods: It will retry every 15 seconds until it works.

someword (approved):

controllers/humioparser_controller.go
someword: thanks!

SaaldjorMike: @someword Here's a bit of context for how this works:
- If `Reconcile()` returns an error, it gets logged out and it gets requeued immediately.
- If no error is returned, but you still want to requeue immediately you can return `reconcile.Result{Requeue: true}`
- If no error is returned, but you want to requeue after some specified amount of time has passed, you can return `reconcile.Result{RequeueAfter: time.Second * 15}`. Waiting a bit can be helpful in certain cases where there might be a slight delay/race between certain resources becoming available.

There's no knowledge about "previous" runs, so it will just wait 15 seconds and try again no matter how many times the same thing has failed in the past.

SaaldjorMike: If you're curious about where it happens: https://github.com/kubernetes-sigs/controller-runtime/blob/985e819db7af1cd17782ca7ad27644d0ca655054/pkg/internal/controller/controller.go#L296-L322

Notice how the "error" case and the "requeue immediately" both uses `c.Queue.AddRateLimited(req)`, whereas "requeue after X time" is different.

SaaldjorMike (approved):

Issue #574, opened 2022-02-23
Allow overriding the metadata.name field in Helm chart for Deployment


bderrly:
I am using a custom Helm chart which includes this humio-operator Helm chart as a dependency. The Deployment name here is an unfortunate choice as the term `humio` is used all over the place including both the name of the chart I am using and possibly the Helm release name.

I would like to see the option to override the name of the Deployment (or even hard code it to `humio-operator` though this is not as flexible and will cause churn for already deployed installations).

bderrly:
This came about as a result of running the humio-operator in the same kubernetes namespace where the HumioCluster was configured to run as well. The HumioCluster CR is part of an internal Helm chart. The name of the release was `humio` which caused naming conflicts with the resources applied by the operator. I am going to run the operator in a separate namespace which should keep this from being an issue for me going forward but it is still something that should be considered to be a good citizen in the ecosystem. In my (albeit limited thus far) experience with Helm charts many groups allow overriding the name.

PR #577, opened 2022-03-03
Refactor Helm chart resources

SaaldjorMike:
The main change here is to better generate resource names for the
resources created by installing the Helm chart.

This also removes the unnecessary single quote characters around values
in the YAML files, which makes syntax highlighting slightly better for
some editors.

Ideally we'd also change it so that the resource labels would use the
names that can now be overridden, but I'm leaving that out for now.
This also removes the `OPERATOR_NAME` environment variable from the
Deployment resource as we don't use it at all, but perhaps we should
later consider adding it back again and populating it e.g. `fullname`
such that any resources created by this operator installation would
create and only touch resources that it itself installed. This may make
it easier for us to better support running multiple operators on the
same Kubernetes cluster.

Fixes: https://github.com/humio/humio-operator/issues/574

### DIFF
--- a/charts/humio-operator/templates/_helpers.tpl
+++ b/charts/humio-operator/templates/_helpers.tpl
@@ -1,6 +1,29 @@
 {{/*
 Create chart name and version as used by the chart label.
 */}}
-{{- define "humio.chart" -}}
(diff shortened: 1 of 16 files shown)


jswoods (approved):

SaaldjorMike:
Given the amount of conflicts, let's close this one. We probably want to do another round later though to streamline resource names and such.

PR #582, opened 2022-03-11
Allow to change view's description

namm2:
Signed-off-by: Nam Hai Nguyen <namhai.nguyen-ext@commercetools.com>

I'd like to use Humio operator to change a view's description. This action should be similar to changing a repo's description.


jswoods (approved):

jswoods:
This was merged but had to be reverted since there is no description field on the view in the client api. If this is something you still need, please file an issue and I can look at getting the description field added to the view in the api.

PR #590, opened 2022-05-25
Add k8s 1.24 support, remove k8s 1.19 support

jswoods:


someword (approved):
nice. Adding 1.24 support didn't look as bad as I thought it would.

PR #598, opened 2022-06-23
Kenn/add MinReadySeconds to HumioUpdateStrategy

kenn-humio:
I would like some feedback on how necessary it is to add tests with this PR.
If tests are necessary (I assume they are) and can wait, then I will make another PR containing an automated test for this functionality. In that regard I may need some advice on how best to formulate the test.


jswoods (approved):
I think it's safe enough to merge now since we are waiting on it, and add tests later. Though we should file an issue about adding the tests so we don't lose track of it. I'm not exactly sure the right way to test it, but it may involve a new status field to indicate we're waiting for the pod to pass the min ready time.

jswoods:
@kenn-humio - I created a PR in this repo so we can run the tests. One of the checks is failing here https://github.com/humio/humio-operator/runs/7155133165?check_suite_focus=true.

PR #601, opened 2022-07-12
feat: Use Helm 3 best practice for CRD creation

ryanfaircloth:
Helm3 creates CRD records first which can be skipped using --skipCRDs flag the current method used in the helm chart is based on a defunct Helm2 best practice

Signed-off-by: Ryan Faircloth <ryan@dss-i.com>

### DIFF
--- a/charts/humio-operator/templates/crds.yaml
+++ b/charts/humio-operator/templates/crds.yaml
@@ -1,5 +1,3 @@
-{{- if .Values.installCRDs -}}
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -14457,4 +14455,3 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
-{{- end }}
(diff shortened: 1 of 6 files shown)


jswoods:
We would need to follow this up by adjusting our docs here https://library.humio.com/stable/docs/installation/containers/kubernetes/kubernetes-operator-install/. Would this also mean we would not be able to skip CRD creation when using helm v2? In that doc we do have support for helm v2, so if that would no longer work then we'd have to drop support for helm v2.

ryanfaircloth:
> We would need to follow this up by adjusting our docs here https://library.humio.com/stable/docs/installation/containers/kubernetes/kubernetes-operator-install/. Would this also mean we would not be able to skip CRD creation when using helm v2? In that doc we do have support for helm v2, so if that would no longer work then we'd have to drop support for helm v2.

Helm2 is dead ref https://helm.sh/blog/helm-v2-deprecation-timeline/ no updates including security patches and it has real issues. If Anyone is using Helm2 they should aks themselves critical questions.
Technically this is a breaking change so its a major version bump but its also not impacting

jswoods:
This seems reasonable. @SaaldjorMike - any issues with dropping helm v2 support considering it's EOL now.

jswoods (approved):

PR #637, opened 2022-12-23
fix: do not set ZOOKEEPER_URL_FOR_NODE_UUID unless ZOOKEEPER_URL is set

ryanfaircloth:
Starting with Logscale 1.70 we no longer directly depend on zookeeper the vars are deprecated and should only be set when the URL is set to support transition


SaaldjorMike:
Closing due to https://github.com/humio/humio-operator/pull/639 being merged.

PR #640, opened 2023-01-05
Adding crd for humio role

CJNA:
This is a follow up to https://github.com/humio/cli/pull/114/files
Will add a reconciler logic later on top of this


jswoods (commented):
First off, thank you for working on this!

The CRDs are generated after the spec is added. To do this, add the spec as needed and then run `make manifests`. This will write out the CRD(s). I suggest having the spec, CRDs, tests and controller logic in the same PR.

CJNA:
Hey @jswoods 
Thanks for taking a look at this PR

PR #645, opened 2023-01-24
HumioVersionFromString should strip away SHA pin

### LINKED ISSUE #615: HumioVersionFromString fails if image has a sha pin
If you set `image: humio/humio-core:1.34.0@sha256:38c78710107dc76f4f809b457328ff1c6764ae4244952a5fa7d76f6e67ea2390`, then the HumioCluster goes into a ConfigError and fails to start.

I believe this is because `HumioVersionFromString` doesn't properly strip the sha pin portion out of the image tag.

SaaldjorMike:
Fixes #615


ssboisen (approved):
LGTM!

PR #654, opened 2023-02-02
Adjust action tests to run without external DNS resolution

SaaldjorMike:


ssboisen:
demonstrably nasty solution 👍

PR #672, opened 2023-03-08
refactor common labels and new feature set additionnal common labels

roumigus:
1. Refactor common labels with the `_helper.tpl` template
2. Allows us to set extra labels in `commonLabels`. Those labels would be propagated to all k8s objects (as the name implies, they are "labels common to all objects".
3. This does not mess with the `Selector` on the k8s service object so there's no breaking change.

When doing `helm diff` I notice the following:

* there was some generic/common labels missing on the k8s `ServiceMonitor` and `Service` objects. This PR would add them. Here in for the svc:

```diff
logging, humio-operator, Service (v1) has changed:
...
    labels:
      app: 'humio-operator'
      app.kubernetes.io/name: 'humio-operator'
      app.kubernetes.io/instance: 'humio-operator'
+     app.kubernetes.io/managed-by: 'Helm'
+     helm.sh/chart: 'humio-operator-0.18.0'
  spec:
    ports:
```

This is the only non-transparent change, but not a breaking change IMO.

* New feature: if we set the following in our `values.yaml` and deploy, it will add labels everywhere:

`values.yaml`:
```yaml
commonLabels:
  generic_label: everywhere
  foo: bar
```

Helm diff will show the addition of labels. For instance, on the deploy k8s object:
```diff
logging, humio-operator, Deployment (apps) has changed:
...
      app: 'humio-operator'
      app.kubernetes.io/name: 'humio-operator'
      app.kubernetes.io/instance: 'humio-operator'
      app.kubernetes.io/managed-by: 'Helm'
      helm.sh/chart: 'humio-operator-0.18.0'
+     generic_label: everywhere
+     foo: bar
  spec:
    replicas: 1
    strategy:
      type: Recreate
    selector:
...
          app: 'humio-operator'
          app.kubernetes.io/name: 'humio-operator'
          app.kubernetes.io/instance: 'humio-operator'
          app.kubernetes.io/managed-by: 'Helm'
          helm.sh/chart: 'humio-operator-0.18.0'
+         generic_label: everywhere
+         foo: bar
      spec:
        affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
```

### DIFF
--- a/charts/humio-operator/Chart.yaml
+++ b/charts/humio-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: humio-operator
-version: 0.18.0
+version: 0.17.0
 appVersion: 0.17.0
 home: https://github.com/humio/humio-operator
 description: Kubernetes Operator for running Humio on top of Kubernetes
--- a/charts/humio-operator/Chart.yaml
+++ b/charts/humio-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: humio-operator
-version: 0.17.0
+version: 0.18.0
 appVersion: 0.17.0
 home: https://github.com/humio/humio-operator
 description: Kubernetes Operator for running Humio on top of Kubernetes
--- a/charts/humio-operator/templates/operator-rbac.yaml
+++ b/charts/humio-operator/templates/operator-rbac.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.operator.rbac.create -}}
+{{- $commonLabels := include "humio.labels" . }}
 ---
 apiVersion: v1
 kind: ServiceAccount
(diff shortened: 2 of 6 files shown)


charts/humio-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: humio-operator
-version: 0.17.0
+version: 0.18.0
```
roumigus, line 3: Not sure what's the policy regarding bumping the chart version. Are you keeping it in sync with the appVersion bellow?

What we are doing here is adding a new feature in a backward compatible way, so according to semver the we should bump from 0.17.0->0.18.0.

charts/humio-operator/templates/operator-rbac.yaml
```diff
@@ -1,16 +1,13 @@
 {{- if .Values.operator.rbac.create -}}
+{{- $commonLabels := include "humio.labels" . }}
```
roumigus, line 2: Needed because we change the scope with `range` later on (such as in line 12 bellow)

charts/humio-operator/Chart.yaml
jswoods: It's mostly kept in sync, yes. It doesn't need to strictly be the same version of the image but we found it to be confusing if they drift.

I think we'd want to use the version that is present prior to merging this PR. We'll bump this version on the next release.

jswoods (commented):
Hi - thanks for this! The PR looks good with one exception on the chart version. If you can revert that back to 0.17.0, then we can merge this in and I'll bump it to 0.18.0 as part of our next release.

charts/humio-operator/Chart.yaml
roumigus: Sure! I reverted this change: the chart version is now back to `0.17.0`. I will leave the versionning up to you, depending on your release process and schedule.

Issue #671, opened 2023-03-01
feat: Add FieldBased Throttling to Alerts


togie6:
**Summary**
Add support to the humio operator to allow alerts to be created using field based throttling.

**Issue**
I have written a helm chart for the deployment of alerts across repos/views within logscale. Seeing as we have a large number of rules and customers, we need the ability to be able to configure field based throttling within our alerts. Without this, we would have to manually configure each role after deployment. 

Logscale appears to support the ability to create alerts with field based throttling.
<img width="553" alt="image" src="https://user-images.githubusercontent.com/22466611/222012548-dc97ead1-896a-42c7-9b71-0ab7b08e0109.png">

The GraphQL API for [CreateAlert](https://library.humio.com/logscale-graphql-reference/graphql-mutation-field-createalert.html) and [UpdateAlert](https://library.humio.com/logscale-graphql-reference/graphql-mutation-field-updatealert.html) has the ability to create/modify alerts using field based throttling.
<img width="893" alt="image" src="https://user-images.githubusercontent.com/22466611/222012867-3ad0267a-2c65-46bb-81dc-df025a010d04.png">

I noticed that the HumioOperator and Logscale Alert Schema does not support specifying a FieldBasedThrottling.

[Alert CRD](https://github.com/humio/humio-operator/blob/master/charts/humio-operator/crds/core.humio.com_humioalerts.yaml)

[Alert Package Schema](https://schemas.humio.com/alert/v0.2.0)

togie6:
@ryanfaircloth

jswoods:
Fixed by https://github.com/humio/humio-operator/pull/674.

PR #674, opened 2023-03-14
feat: Add support for throttleField to alerts

wodity:
See issue #671


SaaldjorMike:
What happens if the user doesn't specify a `throttleField` in their "HumioAlert" resource?

The `createAlert` GraphQL mutation accepts one of the following from what I can tell:
1. `throttleField` is not specified at all when running the mutation
2. OR, the specified `throttleField` string is non-empty

Just want to make sure we don't break anything if the user doesn't specify `throttleField` as it should behave as option 1 listed above.

wodity:
If you leave it out of the resource, it's not passed to GraphQL due to the `omitempty`. The tests work if it's omitted or included.

SaaldjorMike (approved):

Issue #636, opened 2022-12-20
For Action CR support secretRef


ryanfaircloth:
To safely use action CRs as part of a gitops workflow tokens or other "secrets" should not be part of the definition of the action instead should be placed and used as a SecretRef

Issue #479, opened 2021-11-04
Describing humio action leaks secret value


hoeg:
Hi team,

when creating a `core.humio.com/v1alpha1` resource with a `secretKeyRef` e.g.

```
apiVersion: core.humio.com/v1alpha1
kind: HumioAction
metadata:
  name: "slack-alerts"
spec:
  managedClusterName: humio
  name: "slack-alerts"
  viewName: "services"
  slackPostMessageProperties:
    apiTokenSource:
      secretKeyRef:
        name: slack-secret
        key: slack.token
    channels:
      - "#alerts"
    fields:
      Environment: "{field:environment}"
      Logs: "{events_str}"
```

the value of the `slack-secret` is printed when doing `kubectl describe humioactions.core.humio.com slack-alerts`. Here is a sample output of running describe:

```
...
Spec:
  Managed Cluster Name:  humio
  Name:                  slack-alerts
  Slack Post Message Properties:
    API Token:  <actual value of slack.token>
    API Token Source:
      Secret Key Ref:
        Key:   slack.token
        Name:  slack.token
    Channels:
      #alerts
    Fields:
      Environment:  {field:environment}
      Logs:         {events_str}
  View Name:        services
Status:
  State:  Exists
Events:   <none>
```

I would expect this secret not to be printed. Is this intended behaviour? If not, I wouldn't mind taking a stab at it myself if that is ok?

SaaldjorMike:
Completely agree that it shouldn't show the secret value when describing a `HumioAction` resource. Feel free to take a stab at it.

hoeg:
As there are ongoing development on this repo may I suggest that we create a security advisory for this? I see that there is a fix in the works so it would be great to notify users about this so they can upgrade when time comes.

fanicia:
As a follow-up to this, I had a chat with @hoeg offline and he pointed out that while this issue might seem like a "minor" thing, as the operator is "only" potentially leaking things like Slack or PagerDuty tokens, consider what happens if you make an action that looks like this:


```yaml
piVersion: core.humio.com/v1alpha1
kind: HumioAction
metadata:
  name: "slack-alerts"
spec:
  managedClusterName: humio
  name: "slack-alerts"
  viewName: "services"
  slackPostMessageProperties:
    apiTokenSource:
      secretKeyRef:
        name: prod-aws-key
        key: aws.secret.access.key
    channels:
      - "#alerts"
    fields:
      Environment: "{field:environment}"
[...]

PR #685, opened 2023-04-11
HumioAction: How to improve secret handling?

fanicia:
Both [issue 636](https://github.com/humio/humio-operator/issues/636) as well as [479](https://github.com/humio/humio-operator/issues/479) point out that the current implementation of Humio Action CRs causes secrets to be exposed as plaintext values in the resources.

I had a few hours to look into how this could be improved and I made this PR mostly to spark some discussion. This PR is not meant to be merged as-is. I just wanted to get the maintainer's thoughts on it before going too far in either direction.

# My thoughts on the topic
I just spent a little time looking into it, and it seems like the issue comes from the pattern of resolving secrets into the plaintext fields of the `ha` objects as is done [here](https://github.com/humio/humio-operator/blob/9d939614048952c4dfbf5938eff60ab6ee5cb473/controllers/humioaction_controller.go#L196).

In my opinion, this can be fixed in a couple of ways:

1) Switch pattern so the secret is passed along with the `ha` object down to the point where the HumioClient can reconcile with Humio, such that the `ha` object only ever has a reference to the secret and the secret is never set on the actual `ha` resource. This would mean removing the plaintext fields such as `ha.Spec.SlackPostMessageProperties.ApiToken` from the resources altogether.
2) Make a "hack" where the secrets are removed just before updating the resource's definitions on the kubernetes side. This way, fields like `ha.Spec.SlackPostMessageProperties.ApiToken` never make it into kubernetes, but they can still be used for DTO.

In my opinion 1) would be the cleaner solution. However, it would require some changes in how secrets are handled in the Humio Actions and it would be a breaking change 🙁 It should be noted that this is the recommended way of handling this from a kubernetes point of view.

Option 2) is much less clean, but it would likely be quicker to implement, and it is not a breaking change as it allows people to continue using the plain text versions of the secrets if they so choose.

# What I actually implemented

I wanted to get a feel for what option 2) would look like, even though I don't feel like it is the best solution. As such, I added a hack to `ha.Spec.SlackPostMessageProperties.ApiToken` a few places just before `r.Update` is called. It is definitely the ugly way to solve the issue.

What do the maintainers think? Should something like 2) be made for backward compatibility, should 1) be considered, or am I just completely off base here? 😄

### DIFF
--- a/controllers/humioaction_annotations.go
+++ b/controllers/humioaction_annotations.go
@@ -33,6 +33,7 @@ func (r *HumioActionReconciler) reconcileHumioActionAnnotations(ctx context.Cont
 	}
 
 	// TODO: Hack it here
+	ha.Spec.SlackPostMessageProperties.ApiToken = ""
 	err = r.Update(ctx, actionCR)
 	if err != nil {
 		return reconcile.Result{}, r.logErrorAndReturn(err, "failed to add ID annotation to action")
--- a/controllers/humioaction_annotations.go
+++ b/controllers/humioaction_annotations.go
@@ -32,6 +32,7 @@ func (r *HumioActionReconciler) reconcileHumioActionAnnotations(ctx context.Cont
 		actionCR.ObjectMeta.Annotations[k] = v
 	}
 
+	// TODO: Hack it here
 	err = r.Update(ctx, actionCR)
 	if err != nil {
 		return reconcile.Result{}, r.logErrorAndReturn(err, "failed to add ID annotation to action")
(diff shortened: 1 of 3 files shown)


SaaldjorMike:
I agree with you that having fields with the plaintext secret isn't the way to go, and we should probably stick to only allowing the user to specify secretKeyRef's for the secrets. Not all of the action types currently supports that though, so that needs to be added, and yes, we should probably deprecate/remove the old plaintext `string` type fields holding the plaintext versions when we have a better way for all the action types.

fanicia:
Alright. So it sounds like you would be up for the idea of deprecating the plaintext fields and going for option 1)?
If that's the case, I'll add it to my never-ending list of TODOs and see if I get around to it 🙂 😅

fanicia:
I'll close this PR and reopen a new one if I get around to properly implementing a version of solution 1)

PR #688, opened 2023-04-12
Run storage validation on all node pools and allow to run clusters wi…

SaaldjorMike:
…th pods spawned entirely through nodePools slice

Previously, we only did storage config validation on the HumioCluster "node pool", and not all the node pools. We can also skip storage validation if we don't have a nodeCount > 0.

### DIFF
--- a/examples/humiocluster-nodepool-slice-only.yaml
+++ b/examples/humiocluster-nodepool-slice-only.yaml
@@ -1,0 +1,59 @@
+apiVersion: core.humio.com/v1alpha1
+kind: HumioCluster
+metadata:
+  name: example-humiocluster
+spec:
+  license:
+    secretKeyRef:
+      name: example-humiocluster-license
+      key: data
+  targetReplicationFactor: 2
+  storagePartitionsCount: 720
+  digestPartitionsCount: 720
+  nodeCount: 0
+
+  nodePools:
+    - name: "segments"
+      spec:
+        image: "humio/humio-core:1.76.2"
+        nodeCount: 1
+        extraKafkaConfigs: "security.protocol=PLAINTEXT"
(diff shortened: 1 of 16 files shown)


examples/humiocluster-nodepool-slice-only.yaml
```diff
@@ -0,0 +1,59 @@
...
+  name: example-humiocluster
+spec:
+  license:
+    secretKeyRef:
+      name: example-humiocluster-license
+      key: data
+  targetReplicationFactor: 2
+  storagePartitionsCount: 720
+  digestPartitionsCount: 720
+  nodeCount: 0
```
jswoods, line 13: Should we default to `nodeCount: 0`?

jswoods (approved):

examples/humiocluster-nodepool-slice-only.yaml
SaaldjorMike: I was considering that, but then it would be a breaking change, in the sense that it would then be _required_ to specify `nodeCount: 3` if it isn't explicitly set already. Keeping the default as-is, means this PR doesn't introduce any breaking changes, but solely focusses on making it possible to run purely using the nodePools slice and skip using the "HumioCluster.Spec.NodeCount" pods. Eventually, I'd like us to remove:
```
	HumioNodeSpec `json:",inline"`
```
from here: https://github.com/humio/humio-operator/blob/master/api/v1alpha1/humiocluster_types.go#L89

If we think we should just go ahead and change the default, we can also do that. Just let me know, but I was thinking it would be better to do that separately from this.

SaaldjorMike: I think I will merge this in as-is. Then we can create a followup PR if we want to change the default behavior.

Issue #596, opened 2022-06-14
Support PriorityClassName in pod


ryanfaircloth:
Please add to the CRD and pass a PriorityClassName to the pod template. When using cluster autoscaler in Azure AKS a non default priority is needed to trigger auto scale of the cluster nodes when new pods are added.

PR #689, opened 2023-04-12
Add support for custom PriorityClass on humio pods

jswoods:
Fixes https://github.com/humio/humio-operator/issues/596.

### DIFF
--- a/controllers/humiocluster_defaults.go
+++ b/controllers/humiocluster_defaults.go
@@ -84,6 +84,7 @@ type HumioNodePool struct {
 	path                     string
 	ingress                  humiov1alpha1.HumioClusterIngressSpec
 	clusterAnnotations       map[string]string
+	priorityClassName        string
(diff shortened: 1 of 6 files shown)


controllers/humiocluster_defaults.go
```diff
@@ -197,6 +199,7 @@ func NewHumioNodeManagerFromHumioNodePool(hc *humiov1alpha1.HumioCluster, hnp *h
 			InitServiceAccountName:         hnp.InitServiceAccountName,
 			PodLabels:                      hnp.PodLabels,
 			UpdateStrategy:                 hnp.UpdateStrategy,
+			PriorityClassName:              hc.Spec.PriorityClassName,
```
SaaldjorMike, line 202: I think this should be:
```suggestion
			PriorityClassName:              hnp.PriorityClassName,
```

PR #691, opened 2023-04-19
Add secretRef to PagerDutyProperties

scottevtuch-m1:
Currently only the HumioRepository, OpsGenie, and SlackPostMessage actions support using a secretKeyRef for the API credentials. This PR adds this functionality to the PagerDuty action configuration as well.

PR #692, opened 2023-04-23
HumioActions: Improved secret handling

fanicia:
# OLD PR Notes (Kept for context)
Following my chat with @SaaldjorMike in https://github.com/humio/humio-operator/pull/685 regarding issues [636](https://github.com/humio/humio-operator/issues/636) and [479](https://github.com/humio/humio-operator/issues/479) I have finally sat down and made "the proper solution", which unfortunately is also a breaking change as we discussed.

The code in this PR is accepted by the test suite, but I was not able to get the operator up and running in a kind cluster as described. Whether this is something on my end I am not sure, but I thought I would mention it.

## The approach for the change

Rather than storing secrets in the CRD for the actions `HumioRepository`,  `OpsGenie` and  `slackPostMessage`, I have moved the result from the `resolveSecrets` function to a map to be retrieved later in the `action_transform` logic when it's applied to Humio.

I have removed the fields containing the plaintext versions of the secrets as discussed in my previous PR (this caused a breaking change in the API ⚠️ )

I have also made the secret fields of the CRDs in question required in the yaml-spec as the actions don't make much semantic sense without them.

I also made a small change to `resolveField`, as the function took a `value` parameter that seemed to serve no purpose.

## Final notes

The PR [691](https://github.com/humio/humio-operator/pull/691) currently adds new functionality for a PagerDuty action without removing the previous field. While it's awesome that other people are making these improvements, I suggest also removing the previous field here. Otherwise, issue the issue described in 479 will simply be moved to a new action.
But there are more CRDs with the same issue as the ones I try to solve in this PR.
My thinking was that I would start with the low-hanging fruit from `resolveSecrets` as a beginning, rather than making too many changes at once.

# 2024 notes

Following my interactions with @SaaldjorMike back in April 2023, I wanted to come back to this with a new approach, so we can make the non-breaking changes he suggested. The PR is now in that state!

## The new approach to a solution

* When resolving the `ApiToken` from `SlackPostMessage`, if the cleartext field is set --> use that. If it is not set, take the value from the secret and place the value in a map for later retrieval. This way, secret values are never accidentally placed in the cleartext field like in previous versions.
* In the `humio.CRActionFromAPIAction` function, instead of placing the `ApiToken` in the HumioAction, we also utilize the secretMap.
* During testing in `humioresources_controller_test.go`, I have changed the expected values to verify the value in the map in fitting places.

## TL;DR

I have essentially attempted at point 1 of [Mikes suggestion in April 2023](https://github.com/humio/humio-operator/pull/692#issuecomment-1538190425). If this approach is somewhat in line with how the maintainers want it, I will gladly take a stab at the "bonus points" he mentioned 🙂


SaaldjorMike:
```
this caused a breaking change in the API
```

What does this mean for users that rely on the feature today? How would they be noticing this breaking change? The reason why we've not introduced breaking API changes thus far has been partly due to this exact question. There's multiple things I'd like to restructure in the API but have been generally postponing this for a while now.

It seems like the general recommendation would be to actually bump the CRD version, and then use conversion webhooks to handle the any potential conversions: https://sdk.operatorframework.io/docs/best-practices/best-practices/#summary-1
But I'm not sure how that would actually look like implementation wise with multiple API versions and such, as I've not been digging into that topic yet.

fanicia:
> What does this mean for users that rely on the feature today?

It would mean that existing versions of these CRDs would not work, so yeah I think a version bump is in order, yes.
Full disclosure, I am not an expert in operators and CRDs, so I'll let the strategic decisions on this be up to the maintainers. But if you want me to do anything in this regard, though, let me know. I am happy to change anything to fit this better into the long-term roadmap of the operator.

fanicia:
Concretely, I guess a safe way to do the version bump would be to make my changes here in a `v1alpha2` package and then reconcile those CRDs in addition to the existing ones, rather than making the changes the way I have done here.

The README could then contain a warning that `v1alpha1` is considered deprecated and will be removed in a future version.

SaaldjorMike:
So far we've generally just been adding fields on the existing API versions, with our intention to later spend some cycles on figuring out how we can rip out the cruft / unwanted fields. That way we can keep adding new functionality while not breaking the API.

I just don't know how adding a new API version changes the structure of the operator itself. I definitely have a couple of areas that I'm lacking info on:

- Would we then need to write a conversion webhook or is it possible to skip that? My guess is that we need one.
- How does this change the reconcile logic? Will we have one reconciler per API version, or would we just have one that matches the "stored API version"? My guess is that this means just having one reconciler that works with the "stored API version", and then there'd be conversion webhooks to convert between any API versions we have support for at the same time and the underlying stored API object version.

Anyhow, I just don't know the details on how this would work if we'd bump the API version, as I haven't spent any cycles in to trying to bump the API version. So far, I've only read up on the notes in the operator-sdk/kubebuilder docs.

SaaldjorMike:
@fanicia Right, that's fine. Appreciate the effort. I really do. I mostly raise these things as concerns in this PR to highlight that I've personally not had cycles to look into those areas so far, which is why we've generally been sticking to only adding new things to the API and not breaking things by removing old fields due to missing insights to how that would look like in practice. So from my POV we probably have 2 different directions here.

1. Don't make any breaking changes & stick to current API version, and only add the new field. We could restructure the code in a way such that it doesn't "resolve" the secrets from the sources and put them into the custom resources in plaintext. Instead we should skip "copying" over the sensitive parts into the custom resource and instead just consume it from the secret objects directly. That way describing the custom resources would stop showing the secret if they are referring to a k8s secret.
2. Figure out how we could iterate on the API version. This includes things like the concerns in my last comment https://github.com/humio/humio-operator/pull/692#issuecomment-1520175158

If it were up to me, I think I would go for option 1. That way we stick to not breaking the API and solve the cases for where k8s secrets references are used in the custom resources, while not copying things around. Then we can push the work for the "clean" option which is what option 2 is all about. We have several other places we'd like to iterate on the API, so this could just be added to that pool of work for later when we get some time to deal with it.

SaaldjorMike:
I think I would prefer that we:

1. Refactor the code such that it skips putting the plaintext token into the objects, but perhaps slightly different from the other PR you link to. The PR you link to seem to be removing the token, after it already has been added, and I think I would rather want to make sure we never put the token in there in the first place. Perhaps we can do it such that it is fetched and only kept in-memory while processing the given alert, where the in-memory token/secret can be either the plaintext version given by some user in e.g. `ApiToken` or by fetching it from e.g. `ApiTokenSource`. That way it is easy to later drop the plaintext api token field when we are ready to do that.
2. (bonus points) Update the remaining alert types to have a `VarSource`-like field. Seems like these are still missing: `HumioActionVictorOpsProperties`, `HumioActionSlackProperties` and `HumioActionPagerDutyProperties`. Maybe `HumioActionWebhookProperties` & `HumioActionEmailProperties` could also be updated to do something like this for some fields. This is mostly just to align the different alert types to allow externalizing the secret the same way.

Hopefully this is slightly better. I just want to let the old plaintext field continue to work for a bit longer, such that we can remove that once we have sorted out the "bump API situation", but ensure if someone uses the field like `ApiTokenSource` then that means the token stays only in the source we reference, and is not put into the `HumioAlert` object directly.

fanicia:
Alright, I see your point. Sounds like it should be a middle ground between the two approaches I started. I'll see when I have time to pick this up again 🙂

fanicia:
@SaaldjorMike It took me almost a year, but I finally got around to taking a proper stab at this again. If you read my updated PR statement, I hope it makes sense. If anything needs clarification/rework, let me know 🙂 I should be able to set aside time to handle suggestions so it won't take another year before I have time to look at it again 😅

SaaldjorMike:
Fixed as part of https://github.com/humio/humio-operator/pull/816
@fanicia thank you so much for your contributions to this effort.

PR #711, opened 2023-06-30
Adds HumioUsers CRD

schofield:
Currently a work in progress, soliciting feedback.


schofield:
I fixed the last remaining bug. This is ready for review.

api/v1alpha1/humiouser_types.go
```diff
@@ -0,0 +1,95 @@
...
+	Email string `json:"email,omitempty"`
+	// Company is the compnay of the user
+	Company string `json:"company,omitempty"`
+	// CountryCode is the compnay of the user
+	CountryCode string `json:"countryCode,omitempty"`
+	// Picture is the url to the user's profile picture
+	Picture string `json:"picture,omitempty"`
+	// IsRoot is the root setting for the user
+	IsRoot bool `json:"isRoot,omitempty"`
+	// CreatedAt is date when the user was created
```
jswoods, line 59: Do we need this as part of the spec? I can't think of a reason one would want to set this.

schofield: I have seen people use it and it is configurable so I think it should be in the spec.

bderrly: What controls will there be around the `IsRoot` field? Could anyone push a YAML doc to the API server and give themselves root privileges?

SaaldjorMike: I agree that `CreatedAt` does seem very strange to add (at least to the Spec of the CRD). The go client involved using `Users().Add(...)` doesn't use it (same for `Users().Update(...)`. The only use of it that I see is to reflect back the `CreatedAt` timestamp for a user as for when it was created. This does not make sense in HumioUsers.Spec though. If we really want to ensure our k8s CR's store that data, then we can store it in HumioUsers.Status if we really have to, but I'm not sure we really need it. It definitely doesn't belong in the Spec though, since users cannot specify that themselves, but is given by the user upon user-creation.

pkg/humio/client.go
```diff
@@ -28,6 +28,7 @@ import (
 
 	"github.com/go-logr/logr"
 
+	"github.com/humio/cli/api"
```
SaaldjorMike, line 31: No need to introduce an import that is already there. Notice the line below, we can just reuse that import.
```suggestion
```

pkg/humio/client.go
```diff
@@ -651,3 +660,71 @@ func (h *ClientConfig) GetActionIDsMapForAlerts(config *humioapi.Config, req rec
 	}
 	return actionIdMap, nil
 }
+
+func (h *ClientConfig) AddUser(config *humioapi.Config, req reconcile.Request, hu *humiov1alpha1.HumioUser) (*humioapi.User, error) {
+	user := humioapi.User{Username: hu.Spec.Username}
+	_, err := h.GetHumioClient(config, req).Users().Add(hu.Spec.Username, api.UserChangeSet{
```
SaaldjorMike, line 666:
```suggestion
	_, err := h.GetHumioClient(config, req).Users().Add(hu.Spec.Username, humioapi.UserChangeSet{
```

pkg/humio/client.go
```diff
@@ -651,3 +660,71 @@ func (h *ClientConfig) GetActionIDsMapForAlerts(config *humioapi.Config, req rec
...
+
+	if curUser.Email != hu.Spec.Email ||
+		curUser.FullName != hu.Spec.FullName ||
+		curUser.Company != hu.Spec.Company ||
+		curUser.CountryCode != hu.Spec.CountryCode ||
+		curUser.Picture != hu.Spec.Picture ||
+		curUser.IsRoot != hu.Spec.IsRoot {
+		_, err = h.GetHumioClient(config, req).Users().Update(
+			hu.Spec.Username,
+			api.UserChangeSet{
```
SaaldjorMike, line 707:
```suggestion
			humioapi.UserChangeSet{
```

pkg/humio/client_mock.go
```diff
@@ -215,6 +216,34 @@ func (h *MockClientConfig) DeleteRepository(config *humioapi.Config, req reconci
...
+	h.apiClient.User = humioapi.User{
+		Username:    hu.Spec.Username,
+		ID:          kubernetes.RandomString(),
+		FullName:    hu.Spec.FullName,
+		Email:       hu.Spec.Email,
+		Company:     hu.Spec.Company,
+		CountryCode: hu.Spec.CountryCode,
+		Picture:     hu.Spec.Picture,
+		IsRoot:      hu.Spec.IsRoot,
+		CreatedAt:   hu.Spec.CreatedAt,
```
SaaldjorMike, line 229:
```suggestion
```

SaaldjorMike: This is not used by the AddUserV2 mutation that is called by this, so this doesn't do anything, so no need to have CreatedAt included here.

api/v1alpha1/humiouser_types.go
schofield: That makes sense, I'll remove it.

schofield: > What controls will there be around the `IsRoot` field? Could anyone push a YAML doc to the API server and give themselves root privileges?

Any user that has the ability to create the resource could change the attribute yes but they could also change the cluster resource and change the authentication.

SaaldjorMike:
I'm going to close this for now. Should you pick this up again, feel free to reopen it.

PR #716, opened 2023-08-20
Fix bad error msg in humiocluster_controller.go

JacobValdemar:
Isn't it a typo?


SaaldjorMike:
Yes it is. Good catch. Thanks

SaaldjorMike (approved):

PR #717, opened 2023-08-23
Bootstrap token

jswoods:


api/v1alpha1/humiocluster_types.go
```diff
@@ -118,9 +118,6 @@ type HumioNodeSpec struct {
 	// DataVolumeSource is the volume that is mounted on the humio pods. This conflicts with DataVolumePersistentVolumeClaimSpecTemplate.
 	DataVolumeSource corev1.VolumeSource `json:"dataVolumeSource,omitempty"`
 
-	// AuthServiceAccountName is the name of the Kubernetes Service Account that will be attached to the auth container in the humio pod.
-	AuthServiceAccountName string `json:"authServiceAccountName,omitempty"`
-
```
SaaldjorMike, line 123: What happens to existing custom resources if we rip it out like this?
Depending on the answer to that, we may keep this around, but just mark it as deprecated/unused. Previously I've been keeping the old fields around in our struct's just so that the generated CRD yaml continues to recognize any potential fields that may be set where we then just ignore whatever the value is.

controllers/humiocluster_permission_tokens.go
```diff
@@ -0,0 +1,209 @@
...
+			return "", err
+		}
+		for _, userResult := range allUserResults {
+			if userResult.OrganizationName == "RecoveryRootOrg" {
+				if userResult.SearchMatch == fmt.Sprintf(" | %s () ()", username) {
+					fmt.Printf("Found user ID using multi-organization query.\n")
+					return userResult.EntityId, nil
+				}
+			}
+		}
```
SaaldjorMike, line 39: Some day we should probably refactor this, so that we fetch `subscription` as part of `userResult` and so we can match on `userResult.Subscription = "ClusterOwner"` instead of whatever the name of the cluster owner organization happens to be. I see this is just moving over whatever "user & token management" from the auth sidecar into the Reconciler, so its fine to stick with this right now, but ideally we'd tidy this up a bit.

SaaldjorMike: If single org. mode is active it seems like organization ID is `SINGLE_ORGANIZATION_ID` with `subscription` being `Paying`.

SaaldjorMike (approved):

api/v1alpha1/humiocluster_types.go
jswoods: Added this back into the spec.

SaaldjorMike (approved):

PR #719, opened 2023-08-30
nodeSelector fix

adamajt-l:
[nodeSelector](https://github.com/humio/humio-operator/blob/fac3ed3b75de8f77262dbf5e7efcf5d216825ac0/charts/humio-operator/values.yaml#L24C5-L24C5) is mentioned in values file, but in deployment template nodeSelectors is required. Small fix to unify naming with k8s.


SaaldjorMike (approved):
Good catch. Thank you!

PR #721, opened 2023-08-30
Release operator helm chart 0.19.1

jswoods:


davidylee:
Do you also need to update the values.yaml tag?

davidylee (approved):

charts/humio-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: humio-operator
-version: 0.19.0
+version: 0.19.1
 appVersion: 0.19.0
```
someword, line 4: does the appVersion value need to be bumped or is that only when doing a major/minor bump?

someword (approved):

charts/humio-operator/Chart.yaml
jswoods: This is only bumped when also releasing a new image. In this case the release is only for the helm chart.

PR #732, opened 2023-09-12
Bump kind, cert-manager & k8s.io dependencies and k8s versions used in e2e workflow

SaaldjorMike:


SaaldjorMike:
```
+ helm install cert-manager jetstack/cert-manager --namespace cert-manager --version v1.12.4 --set installCRDs=true --set 'global.imagePullSecrets[0].name=regcred'
Error: INSTALLATION FAILED: chart requires kubeVersion: >= 1.22.0-0 which is incompatible with Kubernetes v1.21.14
```

jswoods (approved):

PR #734, opened 2023-09-12
Bump kind and k8s versions used in e2e workflow

SaaldjorMike:


SaaldjorMike:
Moving this into https://github.com/humio/humio-operator/pull/732

PR #746, opened 2023-10-17
Allow creating/updating View connections with multiple filters per repo

### LINKED ISSUE #745: Cannot provide multiple query strings for the same repo with `HumioView` CR
Due to an issue with the https://github.com/humio/cli API for updating views, it is currently not possibly to do something like this in a `HumioView` CR:

```yaml
kind: HumioView
spec:
  connections:
  - filter: |
      somestring
    repositoryName: repoOne
  - filter: |
      otherstring
    repositoryName: repoOne
  - filter: |
        astring
    repositoryName: repoTwo
  name: mySpecialView
```

The API will currently only pass the last mention of repoOne on to the GraphQL mutation. I created https://github.com/humio/cli/pull/138 to address the issue of the humio client. Once the client has been updated then humio-operator can utilize the new parameter which is `map[string][]string` rather than `map[string]string`.

bderrly:
View connections need to have a different data type to allow multiple filters for a single repository.

This change requires humio/cli#138 to be merged and a new version of the module created. Fixes #745.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/go-logr/zapr v1.2.4
 	github.com/google/go-cmp v0.5.9
 	github.com/google/martian v2.1.0+incompatible
-	github.com/humio/cli v0.32.3
+	github.com/humio/cli v0.32.4-0.20231025112913-b4dceeab38b6
 	github.com/onsi/ginkgo/v2 v2.12.0
 	github.com/onsi/gomega v1.27.11-0.20230807134635-babe25fc5472
 	github.com/prometheus/client_golang v1.16.0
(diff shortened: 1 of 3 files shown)


bderrly:
I built the latest commit off github.com/humio/cli/api@b4dceeab38b61f908bd1d765de0d84aa8a75a91c. I do not know if we want to use `graphql.String` or a basic `string` since it requires depending on the graphql library. It seems fine as there is already a dependency on it elsewhere.

bderrly:
I tested a build of the operator on a dev k8s cluster with a `HumioExternalCluster` CR pointing to my development instance of LogScale. I was able to modify an existing view to have two connections to the same repository as expected. Are there any other tests that we should be adding or modifying before merging this change?

SaaldjorMike:
```
I do not know if we want to use graphql.String or a basic string since it requires depending on the graphql library. It seems fine as there is already a dependency on it elsewhere.
```

That's seems fine. The dependency is there anyways as a transitive dependency, and having it as a direct dependency is fine for now.

SaaldjorMike:
I'll just kick off the test suite once more, and when tests are green we can merge this in.

PR #761, opened 2023-12-06
feat: Add ARM64 build for operator image

ryanfaircloth:
This change provides for arm64 builds and makes minor changes to the workflow to allow correct builds on forks for the image. Buildable forks and cross platform for helper images is not a goal


SaaldjorMike:
This was merged in as part of https://github.com/humio/humio-operator/pull/762

PR #766, opened 2024-01-09
Allow defining env variables in a way that node pools can override defaults

### LINKED ISSUE #765: Node Pool resources not being created with default environment variables
## Problem
The HumioCluster CRD specifies that the environment variables of node pools will be merged with the default environment variables. In other words, `spec.nodePools.spec.environmentVariables` of the HumioCluster CRD is documented with the following: _"EnvironmentVariables that will be merged with default environment variables then set on the humio container"_ (see [URL for documentation](https://doc.crds.dev/github.com/humio/humio-operator/core.humio.com/HumioCluster/v1alpha1@humio-operator-0.20.0) of the HumioCluster CRD, of Humio Operator v0.20.0).

When specifying node pools, the only environment variables that are applied to the node pool resources (i.e., pods) are the environment variables that are explicitly defined under `spec.nodePools.spec.environmentVariables` of the HumioCluster CRD. There are no additional default environment variables being passed down to the node pool resources. This is not the intended behavior.

## Proposed Solution
The verbiage used to document the `spec.nodePools.spec.environmentVariables` field of the HumioCluster CRD is not exactly clear. The documentation being:  _"EnvironmentVariables that will be merged with default environment variables then set on the humio container"_. Upon reading this, I would expect that the environment variables of the HumioCluster CR are propagated down to the node pools as the node pools are sub-resources of the HumioCluster. This behavior coupled with the documentation of the `spec.nodePools.spec.environmentVariables` field is confusing and misleading as there are no additional environment variables being appended to the list of environment variables besides the environment variables that are explicitly listed in the `spec.nodePools.spec.environmentVariables` field.

The proposed solution is to merge the environment variables specified on the node pools along with the environment variables specified directly on the HumioCluster. In other words, merge the list of environment variables in `spec.nodePools.spec.environmentVariables` and `spec.environmentVariables` fields under the `spec.nodePools.spec.environmentVariables` field.

I believe it makes sense to add this under the `Reconcile` function of the HumioCluster Controller, where the node pools are added to the HumioCluster. See [this snippet](https://github.com/humio/humio-operator/blob/operator-0.20.0/controllers/humiocluster_controller.go#L102-L106) as an example.

### Additional Context
Humio Operator version: v0.20.0
Kubernetes version: v1.25.12-gke.500

bderrly:
A quick hack as proof-of-concept/debugging (meant to resolve #765) as we are working towards splitting Humio node pool types. This is based upon the assumption I took reading the CRD docstring that said "EnvironmentVariables that will be merged with default environment variables then set on the humio container" which I took to mean more or less what this PR is doing. If this is the intended behavior then we should discuss how to properly merge the environment variables in case of conflicts.


SaaldjorMike:
I don't think this is the direction we should take, since we're hoping to eventually rip out nodeCount, environmentVariables, etc. from being specified directly on `HumioCluster.Spec`. The intention is that we want to move towards _only_ allowing those options on `HumioCluster.Spec.NodePools` to stop the confusion around having those options available on both the "top level" and as part of "node pools". I'll post some more context to the issue https://github.com/humio/humio-operator/issues/765 in a few minutes.

controllers/humiocluster_defaults.go
```diff
@@ -382,6 +379,10 @@ func (hnp HumioNodePool) GetEnvironmentVariables() []corev1.EnvVar {
 		},
 	}
 
+	for _, defaultEnvVar := range envDefaults {
```
bderrly, line 382: It seems to me that this append loop should be present, otherwise why create the envDefaults variable above? Was it missed before or am I misunderstanding something here?

controllers/suite/clusters/humiocluster_controller_test.go
```diff
@@ -2150,11 +2274,6 @@ var _ = Describe("HumioCluster Controller", func() {
 			}
 			toCreate := suite.ConstructBasicSingleNodeHumioCluster(key, true)
 			toCreate.Spec.Image = oldSupportedHumioVersion
-			// ZOOKEEPER_URL gets filtered out by default in the call to ConstructBasicSingleNodeHumioCluster, so we add it back here
```
bderrly, line 2153: This seems to no longer be true. While testing I found that there were duplicate "ZOOKEEPER_URL" environment variables until I removed this chunk.

api/v1alpha1/humiocluster_types.go
```diff
@@ -208,7 +213,11 @@ type HumioNodeSpec struct {
 	// to the Humio pods
 	HumioServiceLabels map[string]string `json:"humioServiceLabels,omitempty"`
 
-	// EnvironmentVariables that will be merged with default environment variables then set on the humio container
+	// EnvironmentVariables is the set of variables that will be supplied to all Pods.
+	// This set is merged with fallback environment variables (for defaults in case they are not supplied in the Custom Resource),
+	// and spec.CommonEnvironmentVariables (for variables that should be applied to Pods of all node types).
+	// Precedence is given to more environment-specific variables; below the predecence is greater moving left to right.
+	//   spec.environmentVariables -> spec.commonEnvironmentVariables -> spec.nodePools[].environmentVariables
```
SaaldjorMike, line 220: This is _close_ but not quite what I meant. Sorry for that, I likely didn't explain it very well.

I'm looking for us to essentially do something like this for `Spec.EnvironmentVariables` (which is going away eventually):
```go
spec.environmentVariables = mergeEnvVars(hc.Spec.CommonEnvironmentVariables,hc.Spec.EnvironmentVariables)
```
and for `Spec.NodePools` it would be roughly something along these lines:
```go
spec.nodePools[].environmentVariables = mergeEnvVars(hc.Spec.CommonEnvironmentVariables,spec.nodePools[].environmentVariables)
```
in both cases, they can inherit what is in `hc.Spec.CommonEnvironmentVariables`, but we don't want `spec.nodePools[].environmentVariables` to inherit what comes from `hc.Spec.EnvironmentVariables`.

So the changes to `controllers/humiocluster_controller.go` which does this:
```go
// Prefer CommonEnvironmentVariables over top-level HumioCluster.HumioNodeSpec.EnvironmentVariables as the latter will be deprecated in the future.
hc.Spec.CommonEnvironmentVariables = mergeEnvVars(hc.Spec.EnvironmentVariables, hc.Spec.CommonEnvironmentVariables)
```
should be removed. I tried doing so locally and ran the tests, and looks like the test `Humio Cluster Update Environment Variable Multi Node Pool` needs updating to test for this new behavior.

bderrly: This makes sense. I did this thinking it would make the transition to remove the top-level `HumioNodeSpec` easier. However, now that I think about it more, there could be a use case where a cluster has the "default" node pool with some specific envvar that should not be propagated to the other custom node pools.

SaaldjorMike:
@bderrly Can I get you to remove the outdated `controller-gen` and regenerate manifest files using `make manifests`? As far as I can tell, thats the only thing that's currently failing here according to https://github.com/humio/humio-operator/pull/778
I tried doing this myself in as part of https://github.com/humio/humio-operator/pull/778 and build was green.

PR #771, opened 2024-02-09
Deprecate certain fields in HumioCluster and remove use of unused env…

SaaldjorMike:
…ironment variables

Previously the operator would automatically append these to the environment variables for pods, but as of LogScale 1.89.0 they are no longer needed/used and we can use the default behavior. To override default behavior it is possible to do that by referring to the LogScale documentation and setting the relevant environment variables.

The environment variables no longer set for pods running LogScale 1.89.0 and newer:
- DIGEST_REPLICATION_FACTOR
- STORAGE_REPLICATION_FACTOR
- DEFAULT_PARTITION_COUNT

Since this changes the environment variables list for pods if running LogScale 1.89.0+ it means cluster pods will be replaced when upgrading to an operator version with these changes.


SaaldjorMike:
Moved into https://github.com/humio/humio-operator/pull/792

PR #778, opened 2024-04-08
Bderrly/nodepool envs

SaaldjorMike:


api/v1alpha1/humiocluster_types.go
```diff
@@ -208,7 +213,11 @@ type HumioNodeSpec struct {
 	// to the Humio pods
 	HumioServiceLabels map[string]string `json:"humioServiceLabels,omitempty"`
 
-	// EnvironmentVariables that will be merged with default environment variables then set on the humio container
+	// EnvironmentVariables is the set of variables that will be supplied to all Pods.
+	// This set is merged with fallback environment variables (for defaults in case they are not supplied in the Custom Resource),
+	// and spec.CommonEnvironmentVariables (for variables that should be applied to Pods of all node types).
+	// Precedence is given to more environment-specific variables; below the predecence is greater moving left to right.
+	//   spec.environmentVariables -> spec.commonEnvironmentVariables -> spec.nodePools[].environmentVariables
```
SaaldjorMike, line 220: This is _close_ but not quite what I meant. Sorry for that, I likely didn't explain it very well.

I'm looking for us to essentially do something like this for `Spec.EnvironmentVariables` (which is going away eventually):
```go
spec.environmentVariables = mergeEnvVars(hc.Spec.CommonEnvironmentVariables,hc.Spec.EnvironmentVariables)
```
and for `Spec.NodePools` it would be roughly something along these lines:
```go
spec.nodePools[].environmentVariables = mergeEnvVars(hc.Spec.CommonEnvironmentVariables,spec.nodePools[].environmentVariables)
```
in both cases, they can inherit what is in `hc.Spec.CommonEnvironmentVariables`, but we don't want `spec.nodePools[].environmentVariables` to inherit what comes from `hc.Spec.EnvironmentVariables`.

So the changes to `controllers/humiocluster_controller.go` which does this:
```go
// Prefer CommonEnvironmentVariables over top-level HumioCluster.HumioNodeSpec.EnvironmentVariables as the latter will be deprecated in the future.
hc.Spec.CommonEnvironmentVariables = mergeEnvVars(hc.Spec.EnvironmentVariables, hc.Spec.CommonEnvironmentVariables)
```
should be removed. I tried doing so locally and ran the tests, and looks like the test `Humio Cluster Update Environment Variable Multi Node Pool` needs updating to test for this new behavior.

PR #809, opened 2024-06-25
Allow tls without the need for certmanager to be enabled

yenoromm:
Right now if cert manager is set to false even with tls enabled for the humiocluster, the humio-auth container fails to connect to the humio container with the following error indicating it is trying to use http to a https port:
`Got err trying to obtain user ID of admin user: Post http://<pod ip>:8080/graphql: EOF`

This should return true if cert manager is disabled but tls is enabled, current logic requires both to be true to set HUMIO_NODE_URL to https://$(POD_NAME):8080/

### DIFF
--- a/controllers/humiocluster_defaults.go
+++ b/controllers/humiocluster_defaults.go
@@ -804,7 +804,7 @@ func (hnp *HumioNodePool) TLSEnabled() bool {
 		return helpers.UseCertManager()
 	}
 
-	return helpers.UseCertManager() && *hnp.tls.Enabled
+	return helpers.UseCertManager() || *hnp.tls.Enabled
 }
 
 func (hnp *HumioNodePool) GetProbeScheme() corev1.URIScheme {


SaaldjorMike:
Can you please provide some details as to what scenario this solves?
Specifically: How would you do TLS without cert-manager?

Enabling TLS with cert-manager disabled doesn't make much sense, since humio-operator relies on cert-manager to create certificates for us that is mounted into each container. How would you have TLS enabled without that? Are you planning on manually issuing a bunch of certificates and manually mounting those certificates into the pods?

yenoromm:
Yes, creating the certs, disabling cert manager and mounting the keystores and truststore to the pods. This was the way recommended to us by internal Crowdstrike staff Fred Z and Nilesh K to manage our own CA for Logscale. If you believe this is an anti pattern, we are all ears but this was our solution to getting HUMIO_NODE_URL to use https

SaaldjorMike:
There seems to a bunch of tests failing with this change.

There's a bunch of places in the code that needs updating, since the current assumption is that the only way to get a TLS-enabled LogScale cluster is using cert-manager, at least for the intra-cluster connectivity.

Doing a change like this would also need some test coverage.

yenoromm:
Is there a potential that humioclusters.core.humio.com.spec.tls.caSecretName could achieve what we are trying to do without a change to the operator? We asked Grant S for documentation on the CRDs but was told there isn't any but found this when explaining the CRD. I am assuming I store a CA here the humio operator will use that to create the issuer with cert manager and the certs that are required but that is speculation as we haven't gotten a response through the official support channels.

SaaldjorMike:
humio-operator creates an `Issuer` (from cert-manager) if it doesn't exist: https://github.com/humio/humio-operator/blob/master/controllers/humiocluster_controller.go#L943-L954

This is constructed here: https://github.com/humio/humio-operator/blob/master/controllers/humiocluster_tls.go#L150-L165
and the documentation for such a "CA issuer" is here: https://cert-manager.io/docs/configuration/ca/

Depending on what you're trying to archive this may fit your use case, so perhaps give the cert-manager documentation a look to see if it may fit what you're trying to do.

yenoromm:
Worked as described, only additional features I would be looking at adding are in this section:
https://github.com/humio/humio-operator/blob/master/controllers/humiocluster_tls.go#L187-L219

Seems pretty limited with the DNS names, I was hoping to add the humioclusters.core.humio.com.spec.hostname or even the cluster_name.namespace.svc.cluster.local as a DNS name to hit the service from another namespace

PR #833, opened 2024-08-05
Allow retention change on initial repo creation

brads-au:
I noticed that recently newly created repositories do not have their retention settings set. This appears to be because retention settings are not set during creation, but during the first update cycle, which can't update if the `allowDataDeletion` flag isn't set true.

This change:

- Enable `allowDataDeletion` flag during creating of the repo, this will allow retention to be set.
- Move `AutomaticSearch` above retention, this allows automatic search to be set even if retention change is not permitted.


SaaldjorMike:
This should not be necessary as the `createRepository` mutation does allow specifying a retention during the initial repo creation. So if you are not seeing that, then there must be a bug somewhere and that should get fixed rather than relying on "updating" the repo after the initial creation.

brads-au:
Yeah that's fair enough. I looks like the api package doesn't currently support creating repos with retention options, etc. only updating. That will have to be sorted out first.

PR #837, opened 2024-08-07
Added support for aggregate alerts

triceras:


examples/humioaggregatealert.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: core.humio.com/v1alpha1
+kind: HumioAggregateAlert
+metadata:
+    name: example-aggregate-alert-managed
+spec:
+  managedClusterName: example-humiocluster
+  name: example-aggregate-alert
+  queryString: "#repo = humio | error = true "
```
SaaldjorMike, line 8: Both examples should have an aggregate query as examples, so perhaps append ` | count()` to the query strings.

SaaldjorMike (changes requested):
Awesome work. Mostly minor things, but they do need fixing before it is ready to get merged.

See my comments for various lines.

Also:
- we also have to update the operator-rbac.yaml file in the helm chart to add the new type to the 2 lists of the different CRD's we have
- this seems to throw some errors due to some fields not being set correctly:
```
Reason: 'queryTimestampType' Enum value '' is undefined in enum type 'QueryTimestampType'. Known values are: EventTimestamp, IngestTimestamp.

Reason: 'queryOwnershipType' Enum value '' is undefined in enum type 'QueryOwnershipType'. Known values are: User, Organization.
```
which causes aggregate alert creation to fail. Both filter alerts and scheduled search implementations already set the query ownership type, so can probably do the same here, and then also similar things to the timestamp type.
- finally, we have to decide what we want to do for the versioning. Aggregate alerts was introduced in 1.147.0 and tests currently run with older version than that. I ran a quick test with 1.150.0 using https://github.com/humio/humio-operator/pull/841/commits/78b76b9ef1051134aa36c3fe362dfc99d3ce1068 but that ran into the issue I mentioned about `QueryTimestampType` and `QueryOwnershipType`.

config/samples/core_v1alpha1_humioaggregatealert.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: core.humio.com/v1alpha1
+kind: HumioAggregateAlert
+metadata:
+  labels:
+    name: humioaggregatealert-sample
```
SaaldjorMike, line 5:
```suggestion
  name: humioaggregatealert-sample
```

controllers/humiofilteralert_controller.go
```diff
@@ -218,6 +218,5 @@ func (r *HumioFilterAlertReconciler) logErrorAndReturn(err error, msg string) er
 }
 
 func sanitizeFilterAlert(filterAlert *humioapi.FilterAlert) {
-	filterAlert.ID = ""
```
SaaldjorMike, line 221: This was merged to master, so not sure if it is a merge of master into this branch/rebase gone wrong as it should not show up in this PR. You may need to do the same for aggregate alerts though, but removing this line and the same line for scheduled search is not supposed to be in this PR

pkg/humio/filteralert_transform.go
```diff
@@ -27,6 +27,10 @@ func FilterAlertTransform(hfa *humiov1alpha1.HumioFilterAlert) (*humioapi.Filter
 		filterAlert.ID = hfa.ObjectMeta.Annotations[FilterAlertIdentifierAnnotation]
 	}
 
+	if filterAlert.Labels == nil {
+		filterAlert.Labels = []string{}
+	}
+
 	return filterAlert, nil
```
SaaldjorMike, line 34: Same as the comment from on the changes to sanitizeFilterAlert. This should not be part of this PR. Same goes for updates to pkg/humio/scheduledsearch_transform.go

.github/workflows/e2e.yaml
```diff
@@ -55,4 +55,4 @@ jobs:
         curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.22.0/kind-linux-amd64
         chmod +x ./kind
         ./kind delete cluster || true
-        make clean
+        make clean
```
SaaldjorMike, line 58: You should clean up the state of this PR so that this is not included. It just creates unnecessary noise as this makes no real difference to the core work in this PR. Same goes for the update to .github/workflows/preview.yaml

controllers/humiofilteralert_controller.go
triceras: Not sure why it is showing up on this PR. Let me remove it.

pkg/humio/filteralert_transform.go
triceras: I do not have any idea why it is showing up on this PR. Perhaps I accidently edited the file. I remove it from the PR.

config/samples/core_v1alpha1_humioaggregatealert.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: core.humio.com/v1alpha1
+kind: HumioAggregateAlert
+metadata:
+  labels:
```
SaaldjorMike, line 4:
```suggestion
```
if we dont have any labels, i dont see why we would have that here

pkg/humio/scheduledsearch_transform.go
```diff
@@ -30,6 +30,10 @@ func ScheduledSearchTransform(hss *humiov1alpha1.HumioScheduledSearch) (*humioap
 		scheduledSearch.ID = hss.ObjectMeta.Annotations[ScheduledSearchIdentifierAnnotation]
 	}
 
+	if scheduledSearch.Labels == nil {
+		scheduledSearch.Labels = []string{}
+	}
+
```
SaaldjorMike, line 36: Not related to this PR. Not sure how this got here

SaaldjorMike (approved):
LGTM. Looks like the only failing tests we have is related to the concurrency issue with the mock, but https://github.com/humio/humio-operator/pull/843 will refactor how that works and https://github.com/humio/humio-operator/pull/843 will be updated to fix this for aggregate alerts as well.

PR #849, opened 2024-08-27
Rafael/set helper image

triceras:


charts/humio-operator/templates/operator-deployment.yaml
```diff
@@ -34,6 +34,14 @@ spec:
...
         {{- toYaml . | nindent 8 }}
 {{- end }}
+{{- with .Values.image }}
+      image:
+            {{- toYaml . | nindent 8 }}
+{{- end }}
+{{- with .Values.helperImage }}
+      helperImage:
+            {{- toYaml . | nindent 8 }}
+{{- end }}
```
SaaldjorMike, line 44: What is the purpose with this? Those fields are not valid for a k8s `Deployment` object.

charts/humio-operator/values.yaml
```diff
@@ -40,3 +40,6 @@ operator:
             - linux
 
 certmanager: true
+
+image: humio/humio-core:latest
+helperImage: humio/humio-operator-helper:latest
```
SaaldjorMike, line 45: Please leave the default behavior to use the hardcoded golang versions. We should never have something that defaults to using the `latest` tag.

charts/humio-operator/templates/operator-deployment.yaml
triceras: I just realized that and I made chnages so the HelperImage and Image are passed as environment variables from this file.

charts/humio-operator/templates/operator-deployment.yaml
```diff
@@ -64,6 +64,14 @@ spec:
           value: "humio-operator"
         - name: USE_CERTMANAGER
           value: {{ .Values.certmanager | quote }}
+        {{- if .Values.image }}
+        - name: HUMIO_OPERATOR_DEFAULT_IMAGE
+          value: {{ .Values.image }}
+        {{- end }}
```
SaaldjorMike, line 70: Given that the golang source code already checks if the value is `!= ""` maybe we could simplify this and just do:
```suggestion
        - name: HUMIO_OPERATOR_DEFAULT_IMAGE
          value: {{ .Values.defaultImage | quote }}
```
that makes it read much more in line with the one above. In this case I prefixed the `image` with `default` to make it a bit more explicit as to what it does. Perhaps it should even be `defaultHumioCoreImage`? Simply calling it `image` feels a bit too generic to me and doesn't really say much about what it is when you see it.

One could say the same about the env var `HUMIO_OPERATOR_DEFAULT_IMAGE` as that doesn't say much about what image we're talking about, so perhaps that should be more explicit `HUMIO_OPERATOR_DEFAULT_HUMIO_CORE_IMAGE`? Users are not expected to explicitly set that env var anyway, but rather install the operator with helm, so doesn't matter if the env var name is a little long.

And then we'd probably also need to add an entry in values.yaml where it sits next to `certmanager: ...` with perhaps `defaultImage: ""`. That way the values.yaml does have a mention of this setting being there, whereas this way users wouldn't know unless they go digging through the template. Perhaps we could also include comments in the values.yaml next to the config on what the field does.

charts/humio-operator/templates/operator-deployment.yaml
```diff
@@ -64,6 +64,10 @@ spec:
           value: "humio-operator"
         - name: USE_CERTMANAGER
           value: {{ .Values.certmanager | quote }}
+        - name: HUMIO_OPERATOR_DEFAULT_HUMIO_CORE_IMAGE
+          value: {{ .Values.defaultHumioCoreImage }}
```
SaaldjorMike, line 68:
```suggestion
          value: {{ .Values.defaultHumioCoreImage | quote }}
```

charts/humio-operator/templates/operator-deployment.yaml
```diff
@@ -64,6 +64,10 @@ spec:
...
+        - name: HUMIO_OPERATOR_DEFAULT_HUMIO_HELPER_IMAGE
+          value: {{ .Values.defaultHumioHelperImage }}
```
SaaldjorMike, line 70:
```suggestion
          value: {{ .Values.defaultHumioHelperImage | quote }}
```

controllers/suite/clusters/humiocluster_controller_test.go
```diff
@@ -19,11 +19,12 @@ package clusters
 import (
 	"context"
 	"fmt"
-	cmapi "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	"os"
 	"reflect"
 	"strings"
 
+	cmapi "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+
```
SaaldjorMike, line 27: Not sure why this is here? Seems completely unrelated to this work.

triceras: I do not have any clue why this ended up here as I was not working on this file

SaaldjorMike (approved):

PR #857, opened 2024-09-23
Make RollingUpdateBestEffort behave a bit more like RollingUpdate

bderrly:
It is somewhat surprising behavior, at least to me, that the "best effort" strategy is only considered in the case of a Logscale version update. I think it makes more sense to have it behave like `RollingUpdate` unless there is a version upgrade in which case the version comparison logic can take over. This will allow things like modifications to environment variables which need to restart pods to have a rolling update even if there is no version update at the same time.


SaaldjorMike:
I just merged in https://github.com/humio/humio-operator/pull/985 which effectively addresses this issue. Closing.

PR #860, opened 2024-09-26
Bump default to humio/humio-core:1.153.1 and add EnableZoneAwareness …

SaaldjorMike:
…field to HumioUpdateStrategy

EnableZoneAwareness is true by default.

When EnableZoneAwareness is true:
- pod replacements will be performed one zone at a time.
- if no zone has been decided upon yet, pick the first one from the pods that needs replacing.
- as soon as no more pods are found in the current zone that needs replacing, the zone marker is cleared.

This also refactors the uses of os.Getenv() to helper functions to reduce the number of places we refer to the env var names.


controllers/suite/clusters/humiocluster_controller_test.go
```diff
@@ -4985,4 +4986,136 @@ var _ = Describe("HumioCluster Controller", func() {
 			}, testTimeout, suite.TestInterval).Should(HaveKeyWithValue(kubernetes.NodePoolLabelName, key.Name))
 		})
 	})
+
+	Context("test rolling update with max unavailable absolute value", Label("dummy"), func() {
```
SaaldjorMike, line 4990:
```suggestion
	Context("test rolling update with max unavailable absolute value", Label("envtest", "dummy"), func() {
```

bsocaciu (approved):
Looks good to me.

I'd look into a way of refactoring `versions.go` to be more 'clean' and less repetitive. But that's work for a future PR.

cs-gschofield (approved):

SaaldjorMike:
Merged as part of https://github.com/humio/humio-operator/pull/862

PR #873, opened 2024-11-01
Ensure created pods are bound to PVCs on schedulable nodes.

HackToHell:
Currently if local PVs are used for nvme/local SSDs, the operator will pick up a PV randomly. This makes it such that it only picks up PVCs that are bound to PV only to nodes which are schedulable. 

The main changes are

- Ensures that if a rolling update is being done with one nodepool marked as non schedulable then the new pods don't bind to the existings pvcs pointing to the unschedulable nodes. 
- PVs that are already bound are prioritized over normal ones.

### DIFF
--- a/controllers/humiocluster_controller.go
+++ b/controllers/humiocluster_controller.go
@@ -2092,14 +2092,14 @@ func (r *HumioClusterReconciler) ensurePersistentVolumeClaimsExist(ctx context.C
 	if err != nil {
 		return r.logErrorAndReturn(err, "failed to list pvcs")
 	}
-	filteredPersistentVolumeCalims, err := r.FilterSchedulablePVCs(ctx, foundPersistentVolumeClaims)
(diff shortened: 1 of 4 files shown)


HackToHell:
@SaaldjorMike Please let me know if you would like more cases covered. This doesn't fully cover a lot of cases such as Node labels and spec toleration but that might be out of scope. This would make sure automated upgrades and node replacements work without manual intervention. I have tested it out on our internal cluster.

bderrly:
@SaaldjorMike any chance you could review this? We have this patch running on a locally modified repo. It would be nice to get it merged upstream so we don't have to maintain it while trying to stay up-to-date with changes here.

SaaldjorMike:
I've just merged https://github.com/humio/humio-operator/pull/982 which was based on your contribution. I needed to fix the merge conflicts and ended up with a slightly differently formatted implementation, but the gist of https://github.com/humio/humio-operator/pull/982 is the same as this one.

Closing.

PR #880, opened 2024-12-04
Generate client using GraphQL schema

SaaldjorMike:
Most notably:
- Removes dependency on API client in github.com/humio/cli. Instead we call GraphQL directly from `ClientConfig` using generated Go code based on the GraphQL schema.
- Bump minimum supported LogScale version to 1.130.0.

Other changes:
- Remove unused operator-sdk GitHub Action.
- Moves some exported functions to an internal package as they are not meant for external consumption.
- Bump gosec and staticcheck that is used by CI GitHub Actions workflow.


internal/humio/license.go
```diff
@@ -0,0 +1,31 @@
+package humio
+
+import (
+	"fmt"
+
+	jose "github.com/go-jose/go-jose/v4"
```
bsocaciu, line 6: `jose` alias is not needed here since we're importing `/jwt` as a separate import

internal/api/httpclient.go
```diff
@@ -0,0 +1,121 @@
...
+	headers map[string]string
+}
+
+func NewHttpTransport(config Config) *http.Transport {
+	dialContext := config.DialContext
+	if dialContext == nil {
+		dialContext = (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
```
bsocaciu, line 25: DualStack is no longer needed since Fast Fallback is enabled by default

controllers/suite/common.go
```diff
@@ -528,20 +528,23 @@ func CreateAndBootstrapCluster(ctx context.Context, k8sClient client.Client, hum
 				Expect(clusterConfig).ToNot(BeNil())
 				Expect(clusterConfig.Config()).ToNot(BeNil())
 
-				cluster, err := humioClient.GetClusters(clusterConfig.Config(), reconcile.Request{NamespacedName: key})
+				client := humioClient.GetHumioHttpClient(clusterConfig.Config(), reconcile.Request{NamespacedName: key})
```
bsocaciu, line 531: There is a package name `client` ("sigs.k8s.io/controller-runtime/pkg/client"), consider renaming this to avoid possible collision issues.

controllers/suite/common.go
```diff
@@ -554,17 +557,20 @@ func CreateAndBootstrapCluster(ctx context.Context, k8sClient client.Client, hum
 				Expect(clusterConfig).ToNot(BeNil())
 				Expect(clusterConfig.Config()).ToNot(BeNil())
 
-				cluster, err := humioClient.GetClusters(clusterConfig.Config(), reconcile.Request{NamespacedName: key})
-				if err != nil || len(cluster.Nodes) < 1 {
+				client := humioClient.GetHumioHttpClient(clusterConfig.Config(), reconcile.Request{NamespacedName: key})
```
bsocaciu, line 560: There is a package name `client` ("sigs.k8s.io/controller-runtime/pkg/client"), consider renaming this to avoid possible collision issues.

bsocaciu (changes requested):

bsocaciu (approved):

PR #887, opened 2024-12-13
Clean up the way diffs are printed

SaaldjorMike:


SaaldjorMike:
Replaced by: https://github.com/humio/humio-operator/pull/891

Issue #104, opened 2020-06-19
Add PodDisruptionBudget for humio


jswoods:
We should attach a PodDistruptionBudget to the humio pods: https://kubernetes.io/docs/tasks/run-application/configure-pdb.

How the PDB is configured may depend on how many storage/digest replicas are configured.

PR #900, opened 2025-01-13
Pod Disruption Budget implementation

triceras:
This is a new implementation for Pod Disruption Burget. This implements #104


controllers/humiocluster_controller.go
```diff
@@ -2349,3 +2387,207 @@ func getHumioNodePoolManagers(hc *humiov1alpha1.HumioCluster) HumioNodePoolList
...
+
+	// Construct PDB name, handling cluster-level PDB
+	var pdbName string
+	if nodePool.Name == "" {
+		pdbName = fmt.Sprintf("%s-cluster-pdb", hc.Name) // Unique name for cluster-level PDB
+		r.Log.Info("Building cluster-level PDB", "pdbName", pdbName)
+	} else {
+		pdbName = fmt.Sprintf("%s-%s-pdb", hc.Name, nodePool.Name)
+		r.Log.Info("Building node pool PDB", "pdbName", pdbName, "nodePool", nodePool.Name)
+	}
```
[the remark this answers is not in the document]
SaaldjorMike: Maybe I should add some more context on why it is the way it is regarding "node pools".

Back in the day, we didn't have `HumioCluster.Spec.NodePools` and essentially only had support for running a single set of pods all configured the same. As use-cases demanded it, we refactored out what was specific for a given "set of pods" (aka node pool) from being set directly on HumioCluster.Spec, and moved the fields to the `HumioNodeSpec` type, which is then the still inlined into the `HumioCluster.Spec` for backwards compatibility.

I really hope we'll some day remove this inlining to tidy things up and make it easier to talk about node pools as today it may mean different things depending on the case. Today "node pools" could refer to "just" the `HumioSpec.NodePools` slice or all node pools including the implicit "pool" that comes from the inlining. This may be quite confusing and is why I hope we'll some day only support pod creation through `HumioCluster.Spec.NodePools` slice (like so: https://github.com/humio/humio-operator/blob/master/examples/humiocluster-nodepool-slice-only.yaml) but that would be a breaking change to remove the inlining and is why we've so far punted on this.

controllers/humiocluster_controller.go
```diff
@@ -1922,6 +1928,64 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
...
 
+	pdbSpec := hnp.GetPodDisruptionBudget()
+	r.Log.Info("Entering PDB enforcement check", "nodePool", hnp.GetNodePoolName(), "pdbSpec", pdbSpec, "pdbSpec.Enabled", pdbSpec != nil && pdbSpec.Enabled)
+	if pdbSpec != nil && pdbSpec.Enabled {
+		pdbName := fmt.Sprintf("%s-%s-pdb", hc.Name, hnp.GetNodePoolName())
+		pdb := &policyv1.PodDisruptionBudget{}
+
+		r.Log.Info("Fetching PDB",
+			"pdbName", pdbName,
+			"namespace", hnp.GetNamespace())
```
SaaldjorMike, line 1939: You don't need to add the namespace to the logs yourself as we initialize the logger in the start of the `Reconcile()` function with the namespace: https://github.com/humio/humio-operator/blob/master/controllers/humiocluster_controller.go#L91-L96

With this, it means all logs using `r.Log` would include both the name of the `HumioCluster` resource as well as the namespace of where it is located.

controllers/humiocluster_controller.go
```diff
@@ -1922,6 +1928,64 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
...
+		err := r.Get(ctx, types.NamespacedName{
+			Name:      pdbName,
+			Namespace: hnp.GetNamespace(),
+		}, pdb)
+
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				r.Log.Info("PDB not found for node pool, proceeding without PDB check",
+					"pdb", pdbName,
+					"namespace", hnp.GetNamespace(),
```
SaaldjorMike, line 1950: The same thing is true here where you don't need to explicitly add namespace to the log events, since the logger already includes this to all events.

controllers/humiocluster_controller.go
```diff
@@ -1922,6 +1928,64 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
 
 	podsForDeletion := desiredLifecycleState.podsToBeReplaced
 
+	pdbSpec := hnp.GetPodDisruptionBudget()
+	r.Log.Info("Entering PDB enforcement check", "nodePool", hnp.GetNodePoolName(), "pdbSpec", pdbSpec, "pdbSpec.Enabled", pdbSpec != nil && pdbSpec.Enabled)
+	if pdbSpec != nil && pdbSpec.Enabled {
+		pdbName := fmt.Sprintf("%s-%s-pdb", hc.Name, hnp.GetNodePoolName())
```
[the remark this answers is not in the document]
SaaldjorMike: Instead of constructing `pdbName` here, perhaps you can instead just use `GetPodDisruptionBudgetName()` which already construct the full PDB name.

controllers/humiocluster_controller.go
```diff
@@ -1922,6 +1928,64 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
 
 	podsForDeletion := desiredLifecycleState.podsToBeReplaced
 
+	pdbSpec := hnp.GetPodDisruptionBudget()
+	r.Log.Info("Entering PDB enforcement check", "nodePool", hnp.GetNodePoolName(), "pdbSpec", pdbSpec, "pdbSpec.Enabled", pdbSpec != nil && pdbSpec.Enabled)
```
SaaldjorMike, line 1932: As far as I can tell, we don't really need to add anything to `ensureMismatchedPodsAreDeleted`, since `ensureMismatchedPodsAreDeleted` is handling the case of pod replacements due to config changes and version changes/upgrades.

The case for adding PDB is more about ensuring components "outside the humio-operator" e.g. k8s worker node drain won't bring more pods/instances down than the configured amount. Config changes and version changes/upgrades is already controlled through the configurable update strategy and this is already included in the current master version of `ensureMismatchedPodsAreDeleted`.

Let me know if I missed something. Thank you.

controllers/humiocluster_controller.go
```diff
@@ -2250,6 +2314,13 @@ func (r *HumioClusterReconciler) verifyHumioClusterConfigurationIsValid(ctx cont
 }
 
 func (r *HumioClusterReconciler) cleanupUnusedResources(ctx context.Context, hc *humiov1alpha1.HumioCluster, humioNodePools HumioNodePoolList) (reconcile.Result, error) {
+	if !hc.DeletionTimestamp.IsZero() {
+		if err := r.handlePDBFinalizers(ctx, hc); err != nil {
+			return r.updateStatus(ctx, r.Client.Status(), hc, statusOptions().
+				withMessage(fmt.Sprintf("failed to handle PDB finalizers: %s", err)))
+		}
+	}
```
SaaldjorMike, line 2322: I believe `handlePDBFinalizers` can be removed given that all it does is remove the finalizer. 

As long as `ownerReferences` on the PDB object is configured to point at the `HumioCluster` object, then the PDB's are garbage collected once the `HumioCluster` object is deleted, see https://kubernetes.io/docs/concepts/architecture/garbage-collection/#:~:text=Cascading%20deletion,a%20process%20called%20cascading%20deletion.

controllers/humiocluster_controller.go
```diff
@@ -2274,6 +2345,12 @@ func (r *HumioClusterReconciler) cleanupUnusedResources(ctx context.Context, hc
 				withMessage(err.Error()))
 		}
 	}
+
+	if err := r.cleanupOrphanedPDBs(ctx, hc, &humioNodePools); err != nil {
+		return r.updateStatus(ctx, r.Client.Status(), hc, statusOptions().
+			withMessage(err.Error()))
+	}
+
```
SaaldjorMike, line 2353: The same comment goes here as https://github.com/humio/humio-operator/pull/900/files#r1930459818
Cleanup of PDB objects should already be garbage collected by k8s out of the box as long as owner references is properly set up.

api/v1alpha1/humiocluster_types.go
```diff
@@ -310,6 +312,35 @@ type HumioNodePoolSpec struct {
 	HumioNodeSpec `json:"spec,omitempty"`
 }
 
+// PodDisruptionBudgetSpec defines the desired pod disruption budget configuration
+type HumioPodDisruptionBudgetSpec struct {
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Format=int-or-string
+	// +kubebuilder:validation:Immutable
```
SaaldjorMike, line 319: What is the idea behind making this `Immutable`? I don't see why we would not allow users to change the configs after initially configuring.

api/v1alpha1/humiocluster_types.go
```diff
@@ -310,6 +312,35 @@ type HumioNodePoolSpec struct {
...
+	// MinAvailable is the minimum number of pods that must be available during a disruption.
+	MinAvailable *intstr.IntOrString `json:"minAvailable,omitempty"`
+
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Format=int-or-string
+	// +kubebuilder:validation:Immutable
```
SaaldjorMike, line 325: What is the idea behind making this `Immutable`? I don't see why we would not allow users to change the configs after initially configuring.

controllers/humiocluster_controller.go
```diff
@@ -2349,3 +2426,184 @@ func getHumioNodePoolManagers(hc *humiov1alpha1.HumioCluster) HumioNodePoolList
...
+	if len(pods) == 0 {
+		r.Log.Info("no pods found, skipping PDB creation")
+		return nil
+	}
+
+	desiredPDB, err := r.constructPDB(hc, hnp, pdbSpec)
+	if err != nil {
+		r.Log.Error(err, "failed to construct PDB", "pdbName", hnp.GetPodDisruptionBudgetName(), "namespace", hnp.GetNamespace())
```
SaaldjorMike, line 2451: One more place where you could skip adding the namespace yourself due to the logger already including this in all events.

controllers/humiocluster_controller.go
```diff
@@ -2349,3 +2426,184 @@ func getHumioNodePoolManagers(hc *humiov1alpha1.HumioCluster) HumioNodePoolList
...
+
+	minAvailable := pdbSpec.MinAvailable
+	pdb := &policyv1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:       pdbName,
+			Namespace:  hc.Namespace,
+			Labels:     kubernetes.LabelsForHumio(hc.Name), // Add node pool name label if needed
+			Finalizers: []string{HumioProtectionFinalizer},
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(hc, humiov1alpha1.GroupVersion.WithKind("HumioCluster")),
```
SaaldjorMike, line 2481: Interesting. I wasn't aware of this being a valid way to configure the owner reference.

The other places we use the `controllerutil.SetControllerReference()` function to set the ownerReference. Here's an example of this:
https://github.com/humio/humio-operator/blob/b2e27eaaa2d8879c6cb6e0d2b51b9d4e0aefc769/controllers/humiocluster_controller.go#L414

Do you know which one of these would be the better one to use? Perhaps we should be replacing the use of `controllerutil.SetControllerReference()` with using `metav1.NewControllerRef()`.

controllers/humiocluster_controller.go
```diff
@@ -2349,3 +2426,184 @@ func getHumioNodePoolManagers(hc *humiov1alpha1.HumioCluster) HumioNodePoolList
...
+		}
+	}
+
+	r.Log.V(1).Info("pdb is up-to-date", "pdb", pdbName)
+	return nil
+}
+
+// SemanticPDBsEqual compares two PodDisruptionBudgets and returns true if they are equal
+func SemanticPDBsEqual(desired *policyv1.PodDisruptionBudget, current *policyv1.PodDisruptionBudget) bool {
+	if !equality.Semantic.DeepEqual(desired.Spec.MinAvailable, current.Spec.MinAvailable) {
```
SaaldjorMike, line 2536: Cool. Wasn't aware of this `equality` package. 

I see we have a total of 10 references to `reflect.DeepEqual()` various places in the code base on master branch, and maybe we should consider refactoring those to use `equality.Semantic.DeepEqual()` instead.

We also have 91 places on master where we use `cmp.diff()`. Not all of those actually _need_ the diff so perhaps we could also replace and update those with `equality.Semantic.DeepEqual()`.

I don't expect this to be incorporated in this PR though, just wanted to call it out as a possible cleanup we may consider later as I wasn't aware of this `equality` package.

controllers/humiocluster_controller.go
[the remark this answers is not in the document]
triceras: That's a good observation. However, fwiw I haven't finished refactoring the code

controllers/humiocluster_controller.go
triceras: Thanks for the info.  Actually, I was unsure about the log message structure that and was going to ask about it.  I will keep it simple as `r.Log.Info("Fetching PDB",
			"pdbName", pdbName)`

controllers/humiocluster_controller.go
triceras: Makes total sense you explained. So what I understand what you said is that `ensureMismatchedPodsAreDeleted` don't require explicit PDB handling because the function already coordinates pod replacements using:
- The cluster's update strategy (rolling updates)
- Max unavailable pods calculation
- Readiness checks
- Version/config mismatch detection

And PDB's are primarily for external kubernetes-level disruption, like node drains or maintenance, rather than operator-controlled replacements.

Said that, I have removed the PDB section from `ensureMismatchedPodsAreDeleted`

controllers/humiocluster_controller.go
triceras: Thanks for correcting me. Conditional statement removed.

controllers/humiocluster_controller.go
triceras: I wanted to be extra careful here. Deleted the conditional statement above.

api/v1alpha1/humiocluster_types.go
triceras: Same answer as above

api/v1alpha1/humiocluster_types.go
triceras: I have marked those fields immutable to prevent runtime changes that could break the cluster during updates. However, if the operator can safely handle PDB updates through proper requeuing  and rolling update coordination as you previously said, then we could safely remove the flag.

triceras: I am removing the flag "kubebuilder:validation:Immutable".

controllers/humiocluster_controller.go
triceras: Removed "namespace" form the Log statement.

controllers/humiocluster_controller.go
triceras: Thanks for bringing this up. Had a bit of reflection on this comment. Both approaches are valid but serve different purposes:

1. `controllerutil.SetControllerReference()`  is preferred for most cases because:
- Automatically sets GVK (GroupVersionKind)
- Handles cross-namespace ownership prevention - Provides safety checks for CRD scope compatibility
- Maintains consistency with controller-runtime patterns

2. `metav1.NewControllerRef()` is used here because:

- The PDB is explicitly namespaced with the cluster
- Direct control over owner reference creation was needed
- Avoids potential issues with automatic GVK resolution in some contexts. 

In this case, using using `metav1.NewControllerRef()` directly ensured the owner reference would always have:
```
kind: HumioCluster
```

Even if the type wasn't fully registered in the scheme. However, since we properly registered types in `humiocluster_types.go`  , we can safely use controllerutil.SetControllerReference() which provides better validation.

I would say that using `controllerutil.SetControllerReference()` would be a better choice really. I will switch the PDB related code to use `controllerutil.SetControllerReference()`

controllers/humiocluster_controller.go
triceras: II also would like to tackle those changes on a different PR. Thanks for pointing that out

controllers/humiocluster_controller.go
triceras: Same as above

controllers/humiocluster_controller.go
SaaldjorMike: I see. Thank you a lot for the explanation.

controllers/humiocluster_controller.go
```diff
@@ -2391,3 +2397,100 @@ func getHumioNodePoolManagers(hc *humiov1alpha1.HumioCluster) HumioNodePoolList
...
+
+	return r.createOrUpdatePDB(ctx, hc, hnp, desiredPDB)
+}
+
+// constructPDB creates a PodDisruptionBudget object for a given HumioCluster and HumioNodePool
+func (r *HumioClusterReconciler) constructPDB(hc *humiov1alpha1.HumioCluster, hnp *HumioNodePool, pdbSpec *humiov1alpha1.HumioPodDisruptionBudgetSpec) (*policyv1.PodDisruptionBudget, error) {
+	pdbName := hnp.GetPodDisruptionBudgetName() // Use GetPodDisruptionBudgetName from HumioNodePool
+	// Create and populate labels map
+	labels := kubernetes.LabelsForHumio(hc.Name)
+	labels["humio.com/node-pool"] = hnp.GetNodePoolName()
```
SaaldjorMike, line 2444:
```suggestion
```
I don't think these are used for anything anymore and can removed.

SaaldjorMike (approved):

Issue #179, opened 2020-10-08
Scale down functionality


SaaldjorMike:
At this point we only go through and create pods to match the desired nodeCount, but we do not have anything that will scale down at this point.

bsocaciu:
Completed by #901

PR #901, opened 2025-01-14
Implementing Operator downscaling functionality

bsocaciu:


internal/kubernetes/kubernetes.go
```diff
@@ -24,8 +24,10 @@ import (
 )
 
 const (
-	NodePoolLabelName = "humio.com/node-pool"
-	FeatureLabelName  = "humio.com/feature"
+	NodePoolLabelName        = "humio.com/node-pool"
+	FeatureLabelName         = "humio.com/feature"
+	PodMarkedForDataEviction = "humio.com/marked-for-data-eviction"
+	TopologyKubernetesZone   = "topology.kubernetes.io/zone"
```
SaaldjorMike, line 30: You don't need to create a constant for `TopologyKubernetesZone` but can just use the upstream const like we refer to in the `images/helper/main.go` as `corev1.LabelZoneFailureDomainStable`. It does look like `corev1.LabelZoneFailureDomainStable` was deprecated and replaced by `corev1.LabelTopologyZone` so that's probably the one to use.

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2079,170 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
 
+// Gracefully removes a LogScale pod from the nodepool using the following steps:
+//
+//  1. Matches pod names to node ids
+//  2. Computes the zone from which the pod will be removed base on the current node allocation
+//  3. Iterates through pods and for the first one found in the specified zone, sends an eviction request to the node
+//  4. Checks if the eviction has started (with a timeout of 10 seconds)
+//  5. If the eviction has started, it periodically checks every 60 seconds if the eviction has been completed
+//  6. When the eviction is completed and there is no more data on that node, the node is unregistered from the cluster, and the pod is removed.
+func (r *HumioClusterReconciler) removePodGracefully(ctx context.Context, hc *humiov1alpha1.HumioCluster, podsInNodePool []corev1.Pod, req ctrl.Request) error {
```
SaaldjorMike, line 2099: What is the plan if pods are in a non-working state? Like pods in failed state, or some other state other than `Running`? Also, pods stuck in `Pending` would be a case we'd have to consider.

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2079,170 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+
+			r.Log.Info(fmt.Sprintf("validating node data eviction is in progress for vhost %d", vhost))
+			time.Sleep(time.Second * 1)
+		}
+
+		if !isEvicted {
+			return r.logErrorAndReturn(err, fmt.Sprintf("failed to set eviction status for vhost %d", vhost))
+		}
+
+		pod.Annotations[kubernetes.PodMarkedForDataEviction] = "true"
```
SaaldjorMike, line 2154: Something seems to be missing here. You can't just update the in-memory pod object and expect it to be persisted like this. In order to save the info, you'd have to also call `r.Update()` using the updated object.

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2079,170 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+
+		// poll eviction status
+		pollTick := time.Tick(60 * time.Second)
+		for _ = range pollTick {
```
SaaldjorMike, line 2158: I'm not sure this is really the way to go. If we do it like this, it means the operator would effectively be "locked up" until this loop finishes.
What we typically do is "block for at most X time" in a single`r.Reconcile()` execution. This way there's an upper bound on how long time the operator would be stuck in there.

The reason why I say this is that a single instance of humio-operator may be managing many CRD's, and so we can't have it's behavior to block for too long time because it would block reconciliation of all the other objects.

This is why we generally tend to follow this:
1. do something
2. optional: wait at most X
3. requeue
Then rely on requeue to pick up from where we left off.

SaaldjorMike: The mean idea behind "not doing too much" before requeuing is that the operator may be interrupted at any time, so we can't assume any updates in-memory will survive for long (or even short I guess since it can be restarted at any time).

controllers/humiocluster_controller.go
bsocaciu: I think we should consider `failed` and `pending` pods as part of the active/total number of pod (the current implementation). The idea behind this is that if a pod is in one of those states then we should let it be without acting on it, since if we act on it we might open a whole new can of worms.

For `pending` pods I don't see an use-case where the operator downscaling can affect it. The upscaling does and it's ok for the upscaling to not happen if the pod is pending (current implementation)

For `failed` pods it's even trickier, as we can't evict them, and I wouldn't go as far as to intervene and delete the pod (downscaling flow) or create a new one in its place (upscaling flow). I consider it's better to have it in a failed state for someone to see what went wrong and address the issue instead of trying to replace or fix the pod.

controllers/humiocluster_controller.go
bsocaciu: Fixed ✅

controllers/humiocluster_controller.go
bsocaciu: Fixed ❓ 😅

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2081,228 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
 	return reconcile.Result{}, nil
 }
 
+func (r *HumioClusterReconciler) nodeCanBeSafelyUnregistered(node humiographql.GetEvictionStatusClusterNodesClusterNode) bool {
+	evictionStatus := node.GetEvictionStatus()
+	reasonsNodeCannotBeSafelyUnregistered := node.GetReasonsNodeCannotBeSafelyUnregistered()
+
+	return evictionStatus.GetCurrentlyUnderReplicatedBytes() == 0 &&
+		evictionStatus.GetBytesThatExistOnlyOnThisNode() == 0 &&
+		!evictionStatus.IsDigester &&
```
dominique-brice-cs, line 2149: these 3 lines are redundant with the 3 following lines. I recommend keeping only the `reasonsNodeCannotBeSafelyUnregistered` lines

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2081,228 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+		reasonsNodeCannotBeSafelyUnregistered.GetHasDataThatExistsOnlyOnThisNode() &&
+		reasonsNodeCannotBeSafelyUnregistered.GetHasUnderReplicatedData() &&
+		reasonsNodeCannotBeSafelyUnregistered.GetLeadsDigest()
```
dominique-brice-cs, line 2152: ! Very important !
These three lines should check for `false` value. All 3 booleans must be `false` to safely unregister the node.

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2081,228 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+						return reconcile.Result{}, r.logErrorAndReturn(err, fmt.Sprintf("failed to delete pod %s for vhost %d!", pod.Name, vhost))
+					}
+					for i := 0; i < waitForPodTimeoutSeconds; i++ { // Poll check for unregistering
+						if _, err := r.HumioClient.UnregisterClusterNode(ctx, humioHttpClient, req, vhost, false); err != nil {
+							return reconcile.Result{}, r.logErrorAndReturn(err, fmt.Sprintf("failed to unregister vhost %d", vhost))
+						}
+						humioClusterPrometheusMetrics.Counters.PodsDeleted.Inc()
+					}
+				} else {
+					podsSuccessfullyEvicted++
```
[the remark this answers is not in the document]
bsocaciu: Fixed ✅

controllers/humiocluster_controller.go
bsocaciu: Fixed ✅ , switched to using `RefreshClusterManagementStats`

controllers/humiocluster_controller.go
```diff
@@ -2066,18 +2075,226 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+			}
+		}
+	}
+	// check for pods currently being evicted ---> check the eviction status --> if evicted --> remove node --> else, requeue
 	return reconcile.Result{}, nil
 }
 
+func (r *HumioClusterReconciler) checkEvictionStatusForPod(ctx context.Context, humioHttpClient *humioapi.Client, req ctrl.Request, vhost int) (bool, error) {
+	for i := 0; i < waitForPodTimeoutSeconds; i++ {
+		clusterManagementStatsResponse, err := r.HumioClient.RefreshClusterManagementStats(ctx, humioHttpClient, req, vhost)
```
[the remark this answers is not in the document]
bsocaciu: Fair point, I'll do that :)

controllers/humiocluster_controller.go
```diff
@@ -2066,18 +2075,226 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+				}
+				if nodeCanBeSafelyUnregistered {
+					podsSuccessfullyEvicted++
+					r.Log.Info(fmt.Sprintf("successfully evicted data from vhost %d", vhost))
+					r.Log.Info(fmt.Sprintf("removing vhost %d", vhost))
+					if err := r.Delete(ctx, &pod); err != nil { // Delete pod before unregistering node
+						return reconcile.Result{}, r.logErrorAndReturn(err, fmt.Sprintf("failed to delete pod %s for vhost %d!", pod.Name, vhost))
+					}
+					for i := 0; i < waitForPodTimeoutSeconds; i++ { // Poll check for unregistering
+						if _, err := r.HumioClient.UnregisterClusterNode(ctx, humioHttpClient, req, vhost, false); err != nil {
```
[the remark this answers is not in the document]
bsocaciu: Hm... Do we still need to consider the FeatureF lag if we're checking with `refreshClusterManagementStats` before unregistering? 
I'm thinking of doing the "cheap" query and when `isAlive` is false, I can manually check again with `refreshClusterManagementStats` to avoid the Feature Flag and then unregister the 'old' way (regardless of the feature flag.

bsocaciu:
Fixes issue: #179

api/v1alpha1/humiocluster_types.go
```diff
@@ -58,6 +58,10 @@ type HumioClusterSpec struct {
 	// If all Kubernetes worker nodes are located in the same availability zone, you must set DisableInitContainer to true to use auto rebalancing of partitions.
 	// Deprecated: No longer needed as of 1.89.0 as partitions and segment distribution is now automatically managed by LogScale itself.
 	AutoRebalancePartitions bool `json:"autoRebalancePartitions,omitempty"`
+	// EnableDownscalingFeature (PREVIEW) is a feature flag for enabling the downscaling functionality of the humio operator for this humio cluster.
+	// Default: false
```
SaaldjorMike, line 62: Did you consider also using the kubebuilder tags as is listed here? https://book.kubebuilder.io/reference/markers/crd-validation

bsocaciu: Good point. Fixed ✅

controllers/humiocluster_controller.go
```diff
@@ -2095,18 +2119,348 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+				hc.Status.EvictedNodeIds = append(hc.Status.EvictedNodeIds, vhost) // keep track of the evicted node for unregistering
+				err = r.Status().Update(ctx, hc)
+				if err != nil {
+					r.Log.Error(err, "failed to update cluster status")
+					return reconcile.Result{}, err
+				}
+				r.Log.Info(fmt.Sprintf("removing pod %s containing vhost %d", pod.Name, vhost))
+				if err := r.Delete(ctx, &pod); err != nil { // delete pod before unregistering node
+					return reconcile.Result{}, r.logErrorAndReturn(err, fmt.Sprintf("failed to delete pod %s for vhost %d!", pod.Name, vhost))
+				}
```
SaaldjorMike, line 2193: What happens if the `r.Status().Update(ctx, hc)` call succeeds, but then the `r.Delete(ctx, &pod)` fails? Or perhaps we update status, and then due to the operator being restarted, it doesn't get to do the pod deletion?

SaaldjorMike: If I read the code correctly, it would essentially "just" mean it'll append the ID one more time to the Status, and then try to do the pod deletion again? Perhaps it should only do the append operation if the vhost is not already in the list?

bsocaciu: That's fine. If the delete fails after the update succeeds, then the node will remain marked at Evicted and at the next reconcile run, it will try to delete the pod again.
As per the check at [line 2228](https://github.com/humio/humio-operator/blob/30bcb8eff19572a0e364074dae4bd8896787a4a4/controllers/humiocluster_controller.go#L2228), the node will NOT be unregistered if it's still alive (if the pod is not deleted).

controllers/humiocluster_controller.go
```diff
@@ -2040,10 +2041,14 @@ func (r *HumioClusterReconciler) ingressesMatch(ingress *networkingv1.Ingress, d
...
 
-func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov1alpha1.HumioCluster, hnp *HumioNodePool) (reconcile.Result, error) {
+func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov1alpha1.HumioCluster, hnp *HumioNodePool, req ctrl.Request) (reconcile.Result, error) {
 	// Ensure we have pods for the defined NodeCount.
-	// If scaling down, we will handle the extra/obsolete pods later.
-	foundPodList, err := kubernetes.ListPods(ctx, r, hnp.GetNamespace(), hnp.GetNodePoolLabels())
+
+	// Exclude pods that are currently being evicted --> Ensures K8s keeps track of the pods waiting for eviction and doesn't remove pods continuously
+	labelsToMatch := hnp.GetNodePoolLabels()
+	labelsToMatch[kubernetes.PodMarkedForDataEviction] = "false"
```
jswoods, line 2096: Are we using labels or annotations to manage this?

jswoods: This seems to conflict with other areas that are adding the annotation.

controllers/humiocluster_controller.go
jswoods: Are we using labels or annotations to manage this? This conflicts with the earlier code that is using label.

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2079,170 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+	for _, pod := range podsInNodePool {
+		if pod.Labels[kubernetes.TopologyKubernetesZone] != podRemovalZone {
+			continue
+		}
+		if pod.Spec.NodeName == "" {
+			r.Log.Info(fmt.Sprintf("NodeName is empty for pod %s.", pod.Name))
+			continue
+		}
+		vhost := nodeIdToPodMap[pod.GetName()]
+		err = r.HumioClient.SetIsBeingEvicted(ctx, humioHttpClient, req, vhost, true)
```
jswoods, line 2363: This section makes me nervous. Do we have any protection in preventing runaway evictions?

For example, what if SetIsBeingEvicted returns an error but still executes the eviction, or something else happens (operator restart, etc) during this time after we mark for eviction but before we mark the pod as being evicted?

I think we should do some validation before this step, or change things where we're forced to be a little more careful. For example, in the reconciliation we could:

First check against the humio API to see if any nodes are marked for eviction. If they are, ensure the labels/annotations are set. If it sets the annotation, close the reconciliation loop. If it is already set, then continue.

Second step is to check if there are any pods with the label/annotation and check this against the humio API. If it doesn't match, then trigger the eviction for the pods that have the label and restart the reconciliation loop. If it does match, then continue with matchPodsToHosts() and starting the eviction process for a node assuming a node can be safely evicted.

Add logic to restrict a max number of concurrent evictions by default and possibly offer a way to override via the spec, possibly a max per zone?

This way if someone evicts via the UI then it will eventually lead to the pod being replaced, and we will not inadvertently evict all nodes in the cluster.

controllers/humiocluster_controller.go
```diff
@@ -2074,10 +2079,170 @@ func (r *HumioClusterReconciler) ensurePodsExist(ctx context.Context, hc *humiov
...
+		for _ = range pollTick {
+			nodesStatus, err := r.getClusterNodesStatus(ctx, humioHttpClient, req)
+			if err != nil {
+				return err
+			}
+			readyForRemoval := false
+			for _, node := range nodesStatus {
+				if node.GetId() == vhost {
+					evictionStatus := node.GetEvictionStatus()
+					if evictionStatus.GetTotalSegmentBytes() == 0 {
```
jswoods: Do we actually have to wait for all bytes to be evicted? Or should we wait until all the data exists on at least one other node?

bsocaciu: This is outdated ^^, we're currently checking using the `reasonsNodeCannotBeSafelyUnregistered` return field, which provides boolean checks

controllers/humiocluster_controller.go
bsocaciu: See below

api/v1alpha1/humiocluster_types.go
```diff
@@ -58,6 +58,11 @@ type HumioClusterSpec struct {
...
+	// Preview: this feature is in a preview state
+	//+kubebuilder:default=false
+	EnableDownscalingFeature bool `json:"enableDownscalingFeature,omitempty"`
```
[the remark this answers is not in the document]
bsocaciu: Great idea. I've implemented it, now I'm testing it, along with the eviction comment.

Issue #876, opened 2024-11-06
Configmap <cluster_name>-role-permissions not being reconciled when changing rolePermissions


yenoromm:
Issue:
When changing  rolePermissions for a humiocluster CR after creation, the configmap doesn't get updated.

Expected:
When updating the  rolePermissions for a humiocluster CR, configmap will always be reconciled to the the latest value specified.

Similar issue can be seen with #764

Issue #764, opened 2024-01-02
ensureExtraKafkaConfigsConfigMap does not update ExtraKafkaConfigsConfigMap on change


L1ghtman2k:
https://github.com/humio/humio-operator/blob/8f49054fa3787e9601483575760e639b6589ca4f/controllers/humiocluster_controller.go#L456 function seems to only check if there is already a ExtraKafka ConfigMap, but it does not check the content to ensure consistency with what is inside humiocluster.

Current workaround: Delete the ExtraKafka ConfigMap, and let operator recreate it

PR #906, opened 2025-01-28
Ensure updates to rolePermissions, viewGroupPermissions and extraKafk…

SaaldjorMike:
…aConfigs are reflected in the configmaps

Fixes: https://github.com/humio/humio-operator/issues/764
Fixes: https://github.com/humio/humio-operator/issues/876

### DIFF
--- a/controllers/humiocluster_controller.go
+++ b/controllers/humiocluster_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/humio/humio-operator/internal/kubernetes"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/strings/slices"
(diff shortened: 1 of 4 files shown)


controllers/humiocluster_controller.go
```diff
@@ -447,31 +448,48 @@ func (r *HumioClusterReconciler) validateNodeCount(hc *humiov1alpha1.HumioCluste
 func (r *HumioClusterReconciler) ensureExtraKafkaConfigsConfigMap(ctx context.Context, hc *humiov1alpha1.HumioCluster, hnp *HumioNodePool) error {
 	extraKafkaConfigsConfigMapData := hnp.GetExtraKafkaConfigs()
 	if extraKafkaConfigsConfigMapData == "" {
+		extraKafkaConfigsConfigMap, err := kubernetes.GetConfigMap(ctx, r, hnp.GetExtraKafkaConfigsConfigMapName(), hc.Namespace)
+		if err == nil {
+			// TODO: refactor and move deletion to cleanupUnusedResources
```
bsocaciu, line 453: Is this TODO something it should be done now, or is it a future improvement?

controllers/humiocluster_controller.go
```diff
@@ -537,34 +555,46 @@ func (r *HumioClusterReconciler) ensureViewGroupPermissionsConfigMap(ctx context
 	if viewGroupPermissionsConfigMapData == "" {
 		viewGroupPermissionsConfigMap, err := kubernetes.GetConfigMap(ctx, r, ViewGroupPermissionsConfigMapName(hc), hc.Namespace)
 		if err == nil {
-			if err = r.Delete(ctx, viewGroupPermissionsConfigMap); err != nil {
-				r.Log.Error(err, "unable to delete view group permissions config map")
+			// TODO: refactor and move deletion to cleanupUnusedResources
```
bsocaciu, line 558: Same for this. Also, maybe we could extract the functionality of this if > if into a method as it's duplicate to the `extraKafkaConfigsConfigMapData` one.

Actually, a lot of the `ensureViewGroupPermissionsConfigMap` method logic is similar to the one for KafkaConfigMap ^^

controllers/humiocluster_controller.go
```diff
@@ -575,34 +605,46 @@ func (r *HumioClusterReconciler) ensureRolePermissionsConfigMap(ctx context.Cont
 	if rolePermissionsConfigMapData == "" {
 		rolePermissionsConfigMap, err := kubernetes.GetConfigMap(ctx, r, RolePermissionsConfigMapName(hc), hc.Namespace)
 		if err == nil {
-			if err = r.Delete(ctx, rolePermissionsConfigMap); err != nil {
-				r.Log.Error(err, "unable to delete role permissions config map")
+			// TODO: refactor and move deletion to cleanupUnusedResources
```
bsocaciu, line 608: Same as above

controllers/humiocluster_controller.go
SaaldjorMike: My intention is to keep the code as it is in the current state of the PR/branch, mostly because the rolePermissions and viewGroupPermissions code follow the same style in the code. I've also updated the sections for rolePermissions and viewGroupPermissions to include the same TODO comment. The idea is that we can at a later stage when we refactor one of them, we can refactor all. The diff was just much smaller to "do the same" for extraKafkaConfigs than having to redo rolePermissions and viewGroupPermissions as well.

controllers/humiocluster_controller.go
SaaldjorMike: True, they all follow the same logic for creating/updating/deleting configmaps, but the PR is meant to fix the reported bugs and not so much reduce the code duplication, since we can do that in a followup. That followup could also move the cleanup logic to `cleanupUnusedResources` function as indicated in the comment.

bsocaciu (approved):

PR #907, opened 2025-01-29
Deprecate the ExtraKafkaConfigs field and move cleanup of configmaps …

SaaldjorMike:
…to cleanup function


cs-gschofield (approved):
Looks good to me.

PR #926, opened 2025-02-26
Allow unlimited sizes in HumioRepository

### LINKED ISSUE #925: Unable to create HumioRepositories with unlimited size settings
When attempting to make a new HumioRepository with ingestSizeInGB and storageSizeInGB set to 0 (unlimited) we were receiving the following error:

```one or more objects failed to apply, reason: HumioRepository.core.humio.com "alerts-repo" is invalid: [spec.retention.storageSizeInGB: Invalid value: 0: spec.retention.storageSizeInGB in body should be greater than or equal to 1, spec.retention.ingestSizeInGB: Invalid value: 0: spec.retention.ingestSizeInGB in body should be greater than or equal to 1]```

On closer look it was part of the changes to generate the client using the GraphQL schema as part of the 0.27 release.

Update: Our internal helm chart was setting these values explicitly via a default - so if the values are omitted correctly the repository WILL create as documented.

mcjones-cybera:
Fixes #925 

ingestSizeInGB and storageSizeInGB are currently defined with a minimum of 1. However the value of 0 is required to leave these as unlimited and to only use the timeInDays limits.

Changing this to 0 fixes this issue.

### DIFF
--- a/api/v1alpha1/humiorepository_types.go
+++ b/api/v1alpha1/humiorepository_types.go
@@ -35,10 +35,10 @@ const (
 type HumioRetention struct {
 	// perhaps we should migrate to resource.Quantity? the Humio API needs float64, but that is not supported here, see more here:
 	// https://github.com/kubernetes-sigs/controller-tools/issues/245
-	//+kubebuilder:validation:Minimum=1
+	//+kubebuilder:validation:Minimum=0
 	//+optional
 	IngestSizeInGB *int32 `json:"ingestSizeInGB,omitempty"`
-	//+kubebuilder:validation:Minimum=1
+	//+kubebuilder:validation:Minimum=0
 	//+optional
 	StorageSizeInGB *int32 `json:"storageSizeInGB,omitempty"`
 	//+kubebuilder:validation:Minimum=1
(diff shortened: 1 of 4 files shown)


SaaldjorMike:
Here's the golang code:
https://github.com/humio/humio-operator/blob/master/api/v1alpha1/humiorepository_types.go#L35-L47
after changing that you need to run "make manifests" and "make apidocs" to update the source & CRD docs.

mcjones-cybera:
Thanks a ton for the clarification. I've updated the PR to reflect the correct file and the generated files after running the Make commands and am going to test again.

The docs are clear that omitting it is what we want and thanks to your comment I found the default that was explicitly setting it to 0 instead of leaving it omitted in our code.

I'm going to go ahead and test that omitting works as documented - in which case this change is not explicitly required - but I would suggest that because 0 is an allowed value the change is worth including.

mcjones-cybera:
I can confirm that the documentation is correct and fixing it so the values are omitted is working as expected.

As such - I'm good with this being closed and not merged, but because 0 is a valid value I don't want to close it unilaterally for someone who runs into the same issue we did.

SaaldjorMike:
I'll kick off a round of tests using your changes, and if they pass, I'll very likely merge this in short after.

Thank you again for this.

SaaldjorMike (approved):

PR #935, opened 2025-03-05
Added support for TokenSecretAnnotations for HumioIngestToken

bsocaciu:


SaaldjorMike (changes requested):
It looks like this only works for _new_ secrets. It should also update existing secrets with these annotations. People may start using this feature for existing secrets, or need to change the annotations they use.

SaaldjorMike (approved):

SaaldjorMike (approved):

PR #952, opened 2025-03-22
Update helm to support metrics enablement and port change

### LINKED ISSUE #951: 0.28.1 metrics port not listening
Hi,

It looks like since the kubebuilder v4 work the default metrics settings have changed for the chart/container to be disabled. This also prevents the healthchecks from becoming successful. Also it looks like secure metrics is enabled by default, so it might be worth updating the documentation if this is intended.
https://github.com/humio/humio-operator/blob/b6a79bfeb66d09daf4d34ff6b13c6b5969d50b92/cmd/main.go#L78

~~I haven't had a lot of time to investigate, but configuring the metrics port also results in an error.~~
```
  - command:
    - /manager
    - --metrics-secure=false
    - --metrics-bind-address=":8080"
```

```
{"level":"error","ts":"2025-03-21T10:02:41.862913102Z","caller":"cmd/main.go:351","func":"main.main","msg":"problem running manager","Operator.Commit":"0ebfb17b40b312bacadff4df1f3a68ee70c26722","Operator.Date":"2025-03-18T21:35:54+00:00","Operator.Version":"0.28.1","error":"failed to start metrics server: failed to create listener: listen tcp: lookup tcp/8080\": unknown port","stacktrace":"main.main\n\t/workspace/cmd/main.go:351\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:272"}
```

Edit:

This was just due to quotes, remove them around the bind address.
```
  - command:
    - /manager
    - --metrics-secure=false
    - --metrics-bind-address=:8080
```

Thanks.

brads-au:
This resolves #951, where the operator doesn't become healthy to the API since kubebuilder v4. Added options to enable/disable metrics (enabled by default), and updated probes to use dedicated probe endpoints.

### DIFF
--- a/charts/humio-operator/templates/operator-deployment.yaml
+++ b/charts/humio-operator/templates/operator-deployment.yaml
@@ -53,6 +53,10 @@ spec:
         imagePullPolicy: {{ .Values.operator.image.pullPolicy }}
         command:
         - /manager
+{{- if .Values.operator.metrics.enabled }}
+        - --metrics-bind-address=:{{ .Values.operator.metrics.listen.port }}
+        - --metrics-secure={{ .Values.operator.metrics.secure }}
+{{- end }}
         env:
         - name: POD_NAME
           valueFrom:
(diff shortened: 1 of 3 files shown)


SaaldjorMike (approved):
LGTM. Thank you for this. I completely missed this part when doing the kubebuilder migration work.

PR #955, opened 2025-03-27
Implementing pdf-reader service support 

triceras:


config/rbac/role_binding.yaml
```diff
@@ -10,3 +10,8 @@ subjects:
 - kind: ServiceAccount
   name: controller-manager
   namespace: system
+- kind: ServiceAccount
+  name: humio-operator
+  namespace: logging
+
```
jswoods, line 16: Remove newlines.

jswoods: What caused this to be added to the manager-role?

config/samples/core_v1alpha1_humiopdfrenderservice.yaml
```diff
@@ -0,0 +1,104 @@
...
+  imagePullSecrets:
+    - name: regcred
+  # Number of replicas to run/humio-operator/pkg/humio/humio_tes
+  replicas: 2
+  # Port configuration
+  port: 5123
+  # Service type (matches the Terraform configuration)
+  serviceType: ClusterIP
+  # Environment variables
+  env:
```
jswoods, line 20: I think `env` should be `environmentVariables` to remain consistent with this project.

config/samples/core_v1alpha1_humiopdfrenderservice.yaml
```diff
@@ -0,0 +1,104 @@
+apiVersion: core.humio.com/v1alpha1
```
jswoods, line 1: This file has a lot of comments that are meaningless. I suggest removing them unless they are adding context.

api/v1alpha1/humiopdfrenderservice_types.go
```diff
@@ -0,0 +1,146 @@
...
+	// +optional
+	Port int32 `json:"port,omitempty"`
+
+	// Resources defines the resource requests and limits for the container.
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Env allows to specify environment variables for the service.
+	// +optional
+	Env []corev1.EnvVar `json:"env,omitempty"`
```
jswoods, line 53: I added a similar comment below as well, but I think we should call this `EnvironmentVariables` to remain consistent with this project.

charts/humio-operator/templates/operator-deployment.yaml
```diff
@@ -91,3 +91,7 @@ spec:
           capabilities:
             drop:
             - ALL
+
+
+
+
```
jswoods, line 97: Looks like new lines were added here which are not needed.

charts/humio-operator/templates/operator-rbac.yaml
```diff
@@ -191,6 +194,20 @@ rules:
   - get
   - list
   - watch
+{{- if and .Values.operator.pdfRenderService (eq .Values.operator.pdfRenderService.enabled true) }}
```
jswoods, line 197: We no longer need this conditional since we're not enabling/disabling via the helm chart right?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -0,0 +1,687 @@
...
+
+//+kubebuilder:rbac:groups=core.humio.com,resources=humiopdfrenderservices,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core.humio.com,resources=humiopdfrenderservices/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=core.humio.com,resources=humiopdfrenderservices/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
+
+const (
+	humioPdfRenderServiceFinalizer = "core.humio.com/finalizer"
+	// deploymentName constant removed - using hprs.Name instead
```
jswoods, line 65: I'm unclear on what these comments are referring to? Do we still need them?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -0,0 +1,687 @@
...
+			{Name: "app-temp", MountPath: "/app/temp"},
+			{Name: "tmp", MountPath: "/tmp"},
+		}
+	}
+
+	// Determine container security context, defaulting if necessary
+	containerSecurityContext := hprs.Spec.SecurityContext
+	if containerSecurityContext == nil {
+		containerSecurityContext = &corev1.SecurityContext{
+			AllowPrivilegeEscalation: func() *bool { b := false; return &b }(),
```
jswoods, line 286: We have `helpers.BoolPtr` for these.

internal/controller/suite/resources/humioresources_controller_test.go
```diff
@@ -3832,6 +3838,567 @@ var _ = Describe("Humio Resources Controllers", func() {
...
+		It("should create Deployment and Service when a new HumioPdfRenderService is created", func() {
+			// Create a new HumioPdfRenderService CR with the required fields.
+			cr := &humiov1alpha1.HumioPdfRenderService{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      baseName,
+					Namespace: clusterKey.Namespace,
+				},
+				Spec: humiov1alpha1.HumioPdfRenderServiceSpec{
+					Replicas:           1,
+					Image:              "humio/pdf-render-service:0.0.60--build-102--sha-c8eb95329236ba5fc65659b83af1d84b4703cb1e",
```
jswoods, line 3881: It would be nice if the different image versions were constants.

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -356,22 +421,61 @@ func (r *HumioPdfRenderServiceReconciler) constructDesiredDeployment(hprs *corev
...
+
+	// Ensure limits and requests maps exist
+	if resources.Limits == nil {
+		resources.Limits = corev1.ResourceList{}
+	}
+	if resources.Requests == nil {
+		resources.Requests = corev1.ResourceList{}
+	}
+
+	// Important: If CPU request is 500m for the test case and no limit is set, copy request to limit
```
jswoods, line 435: I'm not sure I understand this. Why not set the limits in the test?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -426,14 +530,59 @@ func normalizeEnvVars(envVars []corev1.EnvVar) []corev1.EnvVar {
 	return result
 }
 
+// internal/controller/humiopdfrenderservice_controller.go
```
jswoods, line 533: Can this be removed?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -446,35 +595,280 @@ func (r *HumioPdfRenderServiceReconciler) getTLSAwareEnvironmentVariables(hprs *
...
+		r.Log.Info("TLS certificate secret missing tls.crt", "secretName", certSecretName)
+	}
+	if _, hasKey := secret.Data["tls.key"]; !hasKey {
+		r.Log.Info("TLS certificate secret missing tls.key", "secretName", certSecretName)
+	}
+
+	return nil
+}
+
+// internal/controller/humiopdfrenderservice_controller.go
```
jswoods, line 645: Same here?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -446,35 +595,280 @@ func (r *HumioPdfRenderServiceReconciler) getTLSAwareEnvironmentVariables(hprs *
...
+
+// configureTLS configures or removes TLS for a deployment based on whether TLS is enabled
+func (r *HumioPdfRenderServiceReconciler) configureTLS(_ context.Context, pdfRenderService *corev1alpha1.HumioPdfRenderService, deployment *appsv1.Deployment) {
```
jswoods, line 648: Why pass in context.Context here?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -187,12 +220,39 @@ func (r *HumioPdfRenderServiceReconciler) Reconcile(ctx context.Context, req ctr
...
+func (r *HumioPdfRenderServiceReconciler) finalize(ctx context.Context, hprs *corev1alpha1.HumioPdfRenderService) error {
+	r.Log.Info("Running finalizer for HumioPdfRenderService", "namespace", hprs.Namespace, "name", hprs.Name)
 
-	// Nothing special to do for this resource as Kubernetes will handle
-	// the garbage collection of owned resources (deployment and service)
-	// This is just a placeholder for any additional cleanup if needed in the future
+	// Explicitly delete the deployment to ensure it's removed
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "pdf-render-service",
```
jswoods, line 230: This should be `r.getResourceName(hprs)`?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -187,12 +220,39 @@ func (r *HumioPdfRenderServiceReconciler) Reconcile(ctx context.Context, req ctr
...
+	err := r.Client.Delete(ctx, deployment)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		r.Log.Error(err, "Failed to delete Deployment during finalization")
+		return err
+	}
+
+	// Explicitly delete the service to ensure it's removed
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "pdf-render-service",
```
jswoods, line 244: This should be `r.getResourceName(hprs)`?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -426,14 +530,59 @@ func normalizeEnvVars(envVars []corev1.EnvVar) []corev1.EnvVar {
...
+	r.Log.Info("Processing TLS environment variables",
+		"CR name", hprs.Name,
+		"TLS enabled", tlsEnabled)
+
+	if tlsEnabled {
+		// Remove any existing TLS variables to prevent duplicates
+		filteredEnvVars := []corev1.EnvVar{}
+		for _, env := range envVars {
+			if env.Name != pdfRenderUseTLSEnvVar &&
+				env.Name != "PDF_RENDER_TLS_CERT_PATH" &&
```
jswoods, line 576: This should be pdfRenderTLSCertPathEnvVar?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -426,14 +530,59 @@ func normalizeEnvVars(envVars []corev1.EnvVar) []corev1.EnvVar {
...
+				env.Name != "PDF_RENDER_TLS_KEY_PATH" {
```
jswoods, line 577: This should be pdfRenderTLSKeyPathEnvVar?

internal/controller/suite/resources/humioresources_controller_test.go
```diff
@@ -4068,120 +4073,125 @@ var _ = Describe("Humio Resources Controllers", func() {
...
+				}
+
+				// Continue with other TLS checks...
 				// Check Volume Mount for tls-cert
 				foundMount := false
 				for _, vm := range container.VolumeMounts {
-					if vm.Name == tlsCertName {
-						g.Expect(vm.MountPath).To(Equal("/etc/ssl/certs/pdf-render-service"), "Incorrect mount path for tls-cert")
-						g.Expect(vm.ReadOnly).To(BeTrue(), "tls-cert mount should be read-only")
+					if vm.Name == "tls-cert" {
```
jswoods, line 4194: These should be tlsCertName?

internal/controller/suite/resources/humioresources_controller_test.go
```diff
@@ -4191,9 +4201,7 @@ var _ = Describe("Humio Resources Controllers", func() {
 				// Check Volume definition for tls-cert
 				foundVolume := false
 				for _, vol := range deployment.Spec.Template.Spec.Volumes {
-					if vol.Name == tlsCertName {
-						g.Expect(vol.VolumeSource.Secret).NotTo(BeNil(), "tls-cert volume source should be a secret")
-						g.Expect(vol.VolumeSource.Secret.SecretName).To(Equal(certSecretName), "tls-cert volume should mount the correct secret")
+					if vol.Name == "tls-cert" {
```
jswoods, line 4204: Same - should be tlsCertName?

internal/controller/humiopdfrenderservice_controller.go
```diff
@@ -446,35 +595,280 @@ func (r *HumioPdfRenderServiceReconciler) getTLSAwareEnvironmentVariables(hprs *
...
 		}
+		return err
+	}
 
-		// Append TLS env vars to existing env vars
-		if envVars == nil {
-			envVars = tlsEnvVars
-		} else {
-			envVars = append(envVars, tlsEnvVars...)
+	// Validate secret has required keys (tls.crt and tls.key)
```
jswoods, line 634: Should we return an error on these?

internal/controller/suite/resources/humioresources_controller_test.go
```diff
@@ -4612,56 +4731,85 @@ var _ = Describe("Humio Resources Controllers", func() {
...
-				// Use deploymentKey which contains the CR name
-				err := k8sClient.Get(ctx, deploymentKey, deployment)
-				if err != nil {
-					return false
+			// Log deployment details to help debug
+			GinkgoT().Logf("Deployment created: %s/%s", deployment.Namespace, deployment.Name)
+			GinkgoT().Logf("Current annotations: %v", deployment.Spec.Template.Annotations)
+
+			// Give the controller more time to update the annotations
+			time.Sleep(5 * time.Second)
```
jswoods, line 4776: A 5 second sleep in the test seems high. Can we just wait on the Eventually()?

jswoods (approved):

api/v1alpha1/humiopdfrenderservice_types.go
```diff
@@ -0,0 +1,161 @@
...
+	HumioPdfRenderServiceStateNotFound = "NotFound"
+	// DefaultPdfRenderServiceLiveness is the default liveness path for the PDF rendering service.
+	DefaultPdfRenderServiceLiveness = "/health"
+	// DefaultPdfRenderServiceReadiness is the default readiness path for the PDF rendering service.
+	DefaultPdfRenderServiceReadiness = "/ready"
+)
+
+// HumioPdfRenderServiceSpec defines the desired state of HumioPdfRenderService
+type HumioPdfRenderServiceSpec struct {
+	// Image is the Docker image to use for the PDF rendering service.
```
SaaldjorMike, line 39:
```suggestion
	// Image is the container image to use for the PDF rendering service.
```

internal/controller/humiocluster_controller.go
```diff
@@ -70,6 +70,8 @@ const (
 	waitingOnPodsMessage = "waiting for pods to become ready"
 
 	humioVersionMinimumForReliableDownscaling = "1.173.0"
+
+	PDFRenderServiceImage = "humio/pdf-render-service:0.0.60--build-102--sha-c8eb95329236ba5fc65659b83af1d84b4703cb1e"
```
SaaldjorMike, line 74: Could to maybe elaborate on what this image is supposed to be? If the goal of this is to be the default image to use if the user didn't specify one, I suggest we put that into `internal/controller/versions/versions.go` where the other default versions are located.

internal/controller/humiocluster_controller.go
```diff
@@ -333,6 +339,47 @@ func (r *HumioClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			// Create new pdf-render-service with TLS copied from cluster
+			newPdfService := &humiov1alpha1.HumioPdfRenderService{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      hc.Name + "-pdf-render-service",
+					Namespace: hc.Namespace,
+					Labels:    map[string]string{"humio_cluster": hc.Name},
+				},
+				Spec: humiov1alpha1.HumioPdfRenderServiceSpec{
+					TLS:      hc.Spec.TLS,
+					Image:    PDFRenderServiceImage,
```
SaaldjorMike, line 360: This one seems to always set a specific value for this. How can users configure the operator to use alternative versions for the render service? We'll have to provide some mechanism so users are able to define which version they'll want to use, especially as bugs may be found we don't want to rely on having to release a new version of "humio-operator" just so that users can use a fixed version of the render service.

internal/controller/humiocluster_controller.go
```diff
@@ -333,6 +339,47 @@ func (r *HumioClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return r.updateStatus(ctx, r.Client.Status(), hc, statusOptions().withState(hc.Status.State).withMessage(""))
 }
 
+// ensurePdfRenderServiceTLS ensures the pdf-render-service resource exists and has TLS config matching the HumioCluster
+func (r *HumioClusterReconciler) ensurePdfRenderServiceTLS(ctx context.Context, hc *humiov1alpha1.HumioCluster) error {
```
SaaldjorMike, line 343: I don't think the name of the function reflects what this actually does. This doesn't have anything to do with TLS as such, but "simply" automatically constructs a `HumioPdfRenderService` object which copies over some of the configs from the "HumioCluster" object. Overall, the primary responsibility of the function is _not_ TLS related, but rather to auto-create the PDF render service object (if I'm reading this right).

internal/controller/humiocluster_controller.go
```diff
@@ -196,6 +198,10 @@ func (r *HumioClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		r.ensureHumioClusterKeystoreSecret,
 		r.ensureNoIngressesIfIngressNotEnabled, // TODO: cleanupUnusedResources seems like a better place for this
 		r.ensureIngress,
+		// Propagate TLS settings to pdf-render-service
+		func(ctx context.Context, hc *humiov1alpha1.HumioCluster) error {
+			return r.ensurePdfRenderServiceTLS(ctx, hc)
+		},
```
SaaldjorMike, line 204: I think the function name and comment here is misleading. It doesn't _just_ propagate TLS configs to the render service object, but straight up also creates it if not already present.

Also, you don't need to do all that `func ...` stuff here. each function listed here is already assumed to have `ctx, hc` as inputs, so you just need to list the function name and the loop will call it:
```suggestion
		r.ensurePdfRenderServiceTLS(ctx, hc)
```

internal/controller/humiocluster_controller.go
SaaldjorMike: Also, the way I read the code, I see no way of disabling this functionality. I can definitely imagine some customers asking for a way of disabling this since they don't need to use the scheduled reports functionality of LogScale.

SaaldjorMike: One more thing. In general, it seems like it'll always create a "minimal" version of `HumioPdfRenderService` which often needs additional configuration tweaks, like affinity rules etc. How would I be able to set those here? Overall it looks like the implementation assumes that:
1. The humio-operator creates a "minimal" version of `HumioPdfRenderService`, which cannot be toggled off.
2. If cluster admins wants to modify it, they'll have to patch/update the object with custom changes after the operator created it (unless they created it themselves beforehand).
3. The operator may be triggering an update of the `HumioPdfRenderService` object to "propagate" some TLS settings.
This overall approach seems to be something that'll very easily run into issues with GitOps-style deployment flows, and we'd need to reconsider this a bit.

SaaldjorMike (changes requested):
I've posted a couple of findings on the PR. I'm mostly concerned about the nature of _always_ creating the `HumioPdfRenderService` object out of the box, with no way for users to disable it, as well as the comments about how this will conflict with GitOps-style deployment workflows.

PR #957, opened 2025-04-05
Added support for feature flags

bsocaciu:


config/samples/core_v1alpha1_humiofeatureflag.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: core.humio.com/v1alpha1
+kind: HumioFeatureFlag
+metadata:
+  name: humiofeatureflag-example
+spec:
+  managedClusterName: example-humiocluster
+#  externalClusterName: example-humiocluster
```
jswoods, line 7: Is this intentional?

internal/controller/suite/resources/humioresources_controller_test.go
```diff
@@ -3374,6 +3374,74 @@ var _ = Describe("Humio Resources Controllers", func() {
...
+				Namespace: clusterKey.Namespace,
+			}
+			toCreateInvalidFeatureFlag := &humiov1alpha1.HumioFeatureFlag{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      key.Name,
+					Namespace: key.Namespace,
+				},
+				Spec: humiov1alpha1.HumioFeatureFlagSpec{
+					ManagedClusterName: clusterKey.Name,
+					//Name: key.Name,
```
jswoods, line 3436: Remove?

internal/controller/suite/resources/humioresources_controller_test.go
```diff
@@ -3374,6 +3374,74 @@ var _ = Describe("Humio Resources Controllers", func() {
...
+				},
+			}
+
+			suite.UsingClusterBy(clusterKey.Name, "HumioFeatureFlag: Trying to create an invalid feature flag")
+			Expect(k8sClient.Create(ctx, toCreateInvalidFeatureFlag)).Should(Not(Succeed()))
```
jswoods, line 3441: Should we also confirm fetchedFeatureFlag.Status.State is in HumioFeatureFlagStateNotFound here?

config/samples/core_v1alpha1_humiofeatureflag.yaml
```diff
@@ -0,0 +1,8 @@
...
+  name: example-featureflag
```
SaaldjorMike, line 8: Maybe we should add an example of a real flag rather than this non-existent flag. I suggest we update it to this:
```suggestion
  name: MultiClusterSearch
```

internal/api/error.go
```diff
@@ -15,6 +15,7 @@ const (
 	entityTypeAction          entityType = "action"
 	entityTypeAlert           entityType = "alert"
 	entityTypeFilterAlert     entityType = "filter-alert"
+	entityTypeFeatureFlag     entityType = "feature-flag"
```
SaaldjorMike, line 18: Interesting. I didn't consider that we'd even add it here. LogScale itself doesn't really treat feature flags as "entities" as far as I'm aware. I wonder if we should consider retiring/removing the entire `internal/api/error.go` file soon, as it is mostly a left-over from the old API client that was located in https://github.com/humio/cli. We only really used this type to wrap our errors and make pretty output when printing errors to users of the `humioctl` tool.

That being said, this is in no way a blocker for this PR, mostly comments that we probably want to get rid of the entire file as it no longer serves the same purpose as it used to.

internal/controller/humiofeatureflag_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		err := r.Update(ctx, featureFlag)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	r.Log.Info("done reconciling")
+	return reconcile.Result{Requeue: true}, nil
```
SaaldjorMike, line 142: I think we should update the implementation to be more in line with the rest of the Reconciler implementations. The other implementations takes a configurable amount of time to requeue rather than what this does, which is requeue immediately.

internal/controller/humiofeatureflag_controller.go
```diff
@@ -0,0 +1,165 @@
...
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	r.Log = r.Log.WithValues("Request.UID", featureFlag.UID)
+
+	cluster, err := helpers.NewCluster(ctx, r, featureFlag.Spec.ManagedClusterName, featureFlag.Spec.ExternalClusterName, featureFlag.Namespace, helpers.UseCertManager(), true, false)
+	if err != nil || cluster == nil || cluster.Config() == nil || !slices.Contains(humiographql.AllFeatureFlag, humiographql.FeatureFlag(featureFlag.Spec.Name)) {
```
SaaldjorMike, line 62: I wonder if we even want to include `|| !slices.Contains(humiographql.AllFeatureFlag, humiographql.FeatureFlag(featureFlag.Spec.Name))` as part of this.
If we do, we would *require* the schema to get updated as new feature flags are introduced, and since the schema is used as a basis for generating the graphql client, it would mean we'd have to release a new operator version and deploy that before we'll be able to toggle a feature flag on/off.
Did we consider if it was an option to fetch this at runtime from the LogScale API so that we'd not have to release&deploy an updated operator to toggle a new feature flag?

internal/api/humiographql/graphql/feature-flags.graphql
```diff
@@ -0,0 +1,19 @@
...
+    $FeatureFlagName: FeatureFlag!
+) {
+    enableFeature(feature: $FeatureFlagName)
+}
+
+mutation DisableFeatureFlag (
+    $FeatureFlagName: FeatureFlag!
+) {
+    disableFeature(feature: $FeatureFlagName)
+}
```
SaaldjorMike, line 19: Feature flags can be enabled in various levels:
- Globally, as in, for an entire cluster
- For a specific organization
- For a specific user

I suggest we rename these grapql mutations and queries slightly so that they are more expressive in that they specifically check for globally-enabled feature flags.

internal/humio/client.go
```diff
@@ -114,6 +115,12 @@ type FilterAlertsClient interface {
 	ValidateActionsForFilterAlert(context.Context, *humioapi.Client, reconcile.Request, *humiov1alpha1.HumioFilterAlert) error
 }
 
+type FeatureFlagsClient interface {
+	EnableFeatureFlag(context.Context, *humioapi.Client, *humiov1alpha1.HumioFeatureFlag) error
+	IsFeatureFlagEnabled(context.Context, *humioapi.Client, *humiov1alpha1.HumioFeatureFlag) (bool, error)
+	DisableFeatureFlag(context.Context, *humioapi.Client, *humiov1alpha1.HumioFeatureFlag) error
```
SaaldjorMike, line 121: Similar to the notes on names for graphql mutations and queries, perhaps we should rename these to more explicitly call out that these feature flags will be enabled "globally" for the entire cluster, just to make them more expressive as LogScale supports enabling feature flags at different levels.

config/samples/core_v1alpha1_humiofeatureflag.yaml
bsocaciu: Yes, it's there to exemplify that one may use both

internal/controller/suite/resources/humioresources_controller_test.go
bsocaciu: I would say to leave it, as per the comment above, it's there to show what is missing in the `invalid test` case

config/samples/core_v1alpha1_humiofeatureflag.yaml
bsocaciu: Good point!

internal/api/error.go
bsocaciu: Hm... makes sense, but should we keep these as "API error wrappers" ? I'm currently using it for and error when a feature flag is not found (said FeatureFlag name doesn't exist in LogScale).
I could also return a custom error for that. Do you think the latter is the right approach?

internal/humio/client.go
bsocaciu: Fixed ✅

internal/api/humiographql/graphql/feature-flags.graphql
bsocaciu: Fixed ✅

internal/controller/humiofeatureflag_controller.go
bsocaciu: Fixed ✅  Fetching those during the reconciliation now. Please check 🙇

internal/controller/humiofeatureflag_controller.go
bsocaciu: Fixed ✅

internal/controller/suite/resources/humioresources_controller_test.go
bsocaciu: Added extra check there. Thanks!

internal/controller/humiofeatureflag_controller.go
SaaldjorMike: I see that it was changed from requeing immediately to "after 15 seconds", but the master branch code currently does it differently. We used to hardcode "requeueafter" to 15 seconds but it is now configurable on the master branch with a default value of 15 seconds as it was before. This was changed in https://github.com/humio/humio-operator/pull/962

So other reconcilers follow this: https://github.com/humio/humio-operator/blob/master/internal/controller/humiofilteralert_controller.go#L190-L191

I suggest we make this small change to use the configurable value for "RequeueAfter" before we merge this, otherwise LGTM.

SaaldjorMike (approved):

Issue #956, opened 2025-04-03
Most resources reconcile every 15 seconds


bderrly:
Why is this the case? Is there some reason that these resources cannot simply be watched and reconciled if they are changed? The default reconcile period is 10 hours (±10% jitter), is that insufficient?

SaaldjorMike:
The exact value of using a default of 15 seconds before running the next reconcile was picked somewhat arbitrarily. This just seemed like a decent value at the time, where the user wouldn't have to sit around waiting for "too long" for the operator to reconcile a change happening outside the operator, like someone using the UI to change settings.

IMO waiting up to 10 hours before the operator would revert a change made by the user outside the operator would be a bit much, but perhaps 15 seconds is a bit too low/aggressive.

Unfortunately we don't have a way to easily watch e.g. parsers through the LogScale API as the operator is able to watch e.g. pods from the Kubernetes API. If we'd have a way to directly "watch" the LogScale resources, then the default would probably be fine, but until then I think the 10 hour default is just too much.

bderrly:
That makes sense, @SaaldjorMike. Have you given thought to having a flag on a resource that prevents modification outside of the operator? For example, if a new parser is created via the operator a flag could be set that does not allow changes in the web interface. Obviously it might require some notification in the web interface that the resource is restricted.

bderrly:
After considering this more I decided to make a PR (#960) as it would solve the problem for our use case. We have roles locked down such that only a few administrators have permission to make changes to resources managed by the humio-operator. This means that we can increase the requeue period without worry about configuration drift.

PR #960, opened 2025-04-15
Add requeue-period flag to humio-operator binary

bderrly:
This flag allows Logscale cluster operators to configure the default requeue period of the Humio* CRs managed by the humio-operator. In deployments where there are sufficient role restrictions it should be possible to prevent competing changes to Logscale entities via the web interface or other GraphQL clients. Thus the requeue period does not need to be such a short period.

Resolves #956

PR #962, opened 2025-04-23
Add requeue-period flag.

SaaldjorMike:
This flag allows operators to manage the default requeue period of the Humio* CRs managed. In deployments where there are sufficient role restrictions it should be possible to prevent competing changes to Logscale entities via the web interface or other GraphQL clients. Thus the requeue period does not need to be such a short period.

Replaces: #960 (this PR takes the original contribution and solves the warning from the linter)
Resolves: #956

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	cmapi "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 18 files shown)


cmd/main.go
```diff
@@ -95,7 +95,8 @@ func main() {
 	flag.StringVar(&metricsCertKey, "metrics-cert-key", "tls.key", "The name of the metrics server key file.")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
-	flag.DurationVar(&requeuePeriod, "requeue-period", 15*time.Second, "The default reconciliation requeue period for all Humio* resources.")
+	flag.DurationVar(&requeuePeriod, "requeue-period", 15*time.Second,
```
bderrly, line 98: One thing I noticed, after sending this off for review, was that I used the <int>*`time.Duration` style whereas the rest of the code uses `time.Duration*<int>` style. Not a big deal but just thought I would call it out in case you want to align it with the rest of the code.

jswoods (approved):

PR #966, opened 2025-04-25
CreateRepository actually uses the retention parameters if provided

bowens-stripe:
# Overview

basically title, actually uses the retention parameters if any is provided from the k8s resource.

I intentionally broke it out to a separate API in the client, `CreateRepositoryWithRetention`, to minimize potential impact to any existing users. It does use the same underlying GraghQL API `createRepository` as the existing one though. LMK if you would prefer to just expand the existing GQL there!

# Testing

`make test` passes locally.

### DIFF
--- a/internal/api/humiographql/graphql/repositories.graphql
+++ b/internal/api/humiographql/graphql/repositories.graphql
@@ -46,6 +46,24 @@ mutation CreateRepository(
     }
 }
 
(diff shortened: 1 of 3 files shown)


SaaldjorMike:
I've kicked off a round of tests. If tests are green then I'll merge it in as-is. Thank you.

SaaldjorMike (approved):

PR #976, opened 2025-05-14
Introduce HumioGroup CRD

SaaldjorMike:


charts/humio-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: humio-operator
-version: 0.28.2
-appVersion: 0.28.2
+version: 0.28.3
+appVersion: 0.28.3
```
bsocaciu, line 4: Shouldn't we do this at release time? The generated charts have `helm.sh/chart: 'humio-operator-0.28.2'`

SaaldjorMike: Oh. Right, I missed this part. Updating now. Correct. Let's keep it at 0.28.2 for now

bsocaciu (approved):

PR #964, opened 2025-04-23
Add support for managing HumioGroup and roles assignment to groups in HumioOperator

bowens-stripe:
# Overview
as title, this PR adds support for basic CRUD of `HumioGroup`s in `HumioOperator` to manage user groups in LogScale, as well as the functionalities to assign and unassign (existing) roles to / from these groups that are associated with (existing) views.

I am using [this PR](https://github.com/humio/humio-operator/pull/257) which added `HumioView` heavily as a reference, as well as existing code for `HumioRepository`

# Test
`make test` passes locally. Please feel free to advice on additional testing strategies as needed!

PR #978, opened 2025-05-15
Introduce CRD's HumioSystemPermissionRole, HumioOrganizationPermissionRole, HumioViewPermissionRole

SaaldjorMike:
This is just the merged result of: https://github.com/humio/humio-operator/pull/974 https://github.com/humio/humio-operator/pull/975 https://github.com/humio/humio-operator/pull/977

I created this single PR as I wanted to deal with merge conflicts _now_. Feel free to review this PR, or the individual ones above. I intend to merge this one and close the other 3.

As shared here https://github.com/humio/humio-operator/pull/964#issuecomment-2878822825 I intend to merge this in, then go work on the PR for adding a `HumioGroup` CRD. When the new `HumioGroup` CRD is there, I'll continue with extending the 3 permission role CRD's with functionality for assigning the role to groups. It's difficult to test assigning them before I have groups to assign them to, which is the reason why the order is like this.

Replaces: https://github.com/humio/humio-operator/pull/974 https://github.com/humio/humio-operator/pull/975 https://github.com/humio/humio-operator/pull/977


internal/humio/client.go
```diff
@@ -1833,6 +1856,86 @@ func (h *ClientConfig) AddUserAndGetUserID(ctx context.Context, client *humioapi
 	}
 }
 
+func (h *ClientConfig) AddSystemPermissionRole(ctx context.Context, client *humioapi.Client, request reconcile.Request, role *humiov1alpha1.HumioSystemPermissionRole) error {
```
bsocaciu, line 1859: request parameter is unused

bsocaciu: same for GetSystemPermissionRole , UpdateSystemPermissionRole , DeleteSystemPermissionRole , AddOrganizationPermissionRole , GetOrganizationPermissionRole , UpdateOrganizationPermissionRole , DeleteOrganizationPermissionRole , AddViewPermissionRole , GetViewPermissionRole , UpdateViewPermissionRole , DeleteViewPermissionRole

SaaldjorMike: Riiiiight. Good call. Actually, this is true for most of the functions we have on the `Client` type. Most of the functions take `reconcile.Request` as input but only a small subset of them actually use it for anything.

So I guess we have a couple of options:
1. Accept the change as-is, then do a cleanup round to tidy up all the functions up where the `reconcile.Request` gets passed as unused input.
2. Update PR to remove it just for these new functions and still do the cleanup round after the fact.
3. Update PR to include removing it not just for the new functions, but for all the functions that doesn't use it to make it go away in all the many places we don't need it.

I'd probably personally lean towards option 1 even though it may not seem ideal to be introducing net new code with unused inputs, but since we have 10+ other types of entities with the same problem, perhaps it'll be fine this time.

Let me know what you prefer. I don't mind whichever of those options we pick.

bsocaciu: I'd prefer to remove it from these new functions as it might create issues where you have to pass arguments through nested calls just to satisfy an unused parameter.
We'll clean up the others afterwards in a separate PR.

bsocaciu (approved):

jswoods (approved):

PR #1024, opened 2025-09-08
Allow Humio Ingest Token to be created without parser name

### LINKED ISSUE #1023: Unable to create Humio Ingest Tokens without a parser
Goal: Create an ingest token without a parser name so the parser within LogScale can be set dynamically.

CRD Usage Example:

    kind: HumioIngestToken
    metadata:
      name: example-humioingesttoken
      namespace: logging
    spec:
      managedClusterName: example-humiocluster
      name: example-humioingesttoken
      repositoryName: example-humiorepository

Expected Result: Creates an ingest token

Instead we are given the error:

    HumioIngestToken.core.humio.com "example-humioingesttoken" is invalid: [spec.parserName: Required value, <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation],

The name of a parser is not supposed to be required in order to create a new ingest token.

Manually editing the installed CRD to remove parserName from the required list, create a working token with no parser set. (See https://github.com/humio/humio-operator/blob/master/config/crd/bases/core.humio.com_humioingesttokens.yaml#L104 for the line deleted)

Documentation links in case it saves someone time:
https://library.humio.com/falcon-logscale-self-hosted/ingesting-data-tokens.html#ingesting-data-tokens-generating-tokens
https://library.humio.com/logscale-graphql-reference-mutations/graphql-mutation-field-addingesttokenv3.html#graphql_example_addingesttokenv3-graphql
https://library.humio.com/humio-operator/installation-containers-kubernetes-operator-resources.html#installation-containers-kubernetes-operator-humioingesttoken

mcjones-cybera:
Fixes #1023

### DIFF
--- a/api/v1alpha1/humioingesttoken_types.go
+++ b/api/v1alpha1/humioingesttoken_types.go
@@ -52,7 +52,6 @@ type HumioIngestTokenSpec struct {
 	Name string `json:"name"`
 	// ParserName is the name of the parser which will be assigned to the ingest token.
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:Required
 	ParserName *string `json:"parserName,omitempty"`
 	// RepositoryName is the name of the Humio repository under which the ingest token will be created
 	// +kubebuilder:validation:MinLength=1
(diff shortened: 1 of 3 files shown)


mcjones-cybera:
This was resolved in 0.33 being released in https://github.com/humio/humio-operator/commit/77b46e773217c30ee61bdf8a91b9fa05

PR #1032, opened 2025-09-24
Adding log level support

bsocaciu:


cmd/main.go
```diff
@@ -98,10 +99,11 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.DurationVar(&requeuePeriod, "requeue-period", 15*time.Second,
 		"The default reconciliation requeue period for all Humio* resources.")
+	flag.StringVar(&logLevel, "loglevel", "INFO", "The level at which to log output.")
```
SaaldjorMike, line 102: We also need somewhere to place the list of possible loglevels we support so that anyone installing the operator knows which ones there are. That could be:
- right here in the flag itself
- in the helm chart default values file
- in the public documentation for installing the operator

Perhaps even multiple places, like the public documentation and the flag we set up here.

PR #1051, opened 2026-04-17
Add expireAfter field to cycle pods after a configured lifetime

xiangyuwang-stripe:
# Summary
- Adds expireAfter (a metav1.Duration) to HumioNodeSpec, configurable per node pool, that defines the maximum time a pod is allowed to run before being replaced
- Implements ensureExpiredPodsAreDeleted in the reconcile loop, which checks pod ages against the TTL and deletes expired pods oldest-first while respecting all existing safety gates (MaxUnavailable, MinReadySeconds, zone awareness, waitingOnPods)
- Runs as a separate reconcile step after spec-drift detection and before pod creation, so image/config updates always take priority over TTL cycling

# Example usage
```
spec:
  expireAfter: "168h"  # cycle pods every 7 days
  # or per node pool:
  nodePools:
    - name: ingest
      spec:
        expireAfter: "24h"
```


internal/controller/humiocluster_controller.go
```diff
@@ -2164,6 +2174,96 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
...
+
+	if podsStatus.waitingOnPods() {
+		r.Log.Info("skipping expireAfter check because not all pods are ready")
+		return emptyResult, nil
+	}
+
+	now := time.Now()
+	maxAge := expireAfter.Duration
+	var expiredPods []corev1.Pod
+	var nearestExpiry time.Duration
```
barkardk, line 2203: Use nearestExpiry.Truncate(time.Second) for clean log output , consider debug 
 r.Log.Info(fmt.Sprintf("no expired pods found, next expiry in %s", nearestExpiry))

internal/controller/humiocluster_controller_test.go
```diff
@@ -150,3 +154,79 @@ func TestGetDuplicateEnvVarsErrorMessage(t *testing.T) {
...
+				ExpireAfter: &d,
+			},
+		}
+		result := hnp.GetExpireAfter()
+		assert.NotNil(t, result)
+		assert.Equal(t, 24*time.Hour, result.Duration)
+	})
+}
+
+func TestPodExpirationFiltering(t *testing.T) {
```
barkardk, line 179: this test does not exercise the production code path. It reimplements the `now.Sub(pod.CreationTimestamp.Time) > maxAge` filter inline (line marked below) and asserts on its own output  ensureExpiredPodsAreDeleted is never called. A refactor of the real  filter could break behavior and this test would still pass.

internal/controller/humiocluster_controller.go
```diff
@@ -2164,6 +2174,96 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
...
+
+	podsStatus, err := r.getPodsStatus(ctx, hc, hnp, foundPodList)
+	if err != nil {
+		return emptyResult, r.logErrorAndReturn(err, "failed to get pod status for expireAfter check")
+	}
+
+	if podsStatus.waitingOnPods() {
+		r.Log.Info("skipping expireAfter check because not all pods are ready")
+		return emptyResult, nil
+	}
```
barkardk, line 2198: waitingOnPods() returns true whenever notReadyCount > 0   

A pod stuck Running-but-NotReady — deadlocked app, unreachable dependency, misconfigured readiness probe — keeps notReadyCount > 0 indefinitely, so this early return blocks expiry forever. 
This guard defeats the feature's main purpose. 
Distinguish an actual error condition (haveUnschedulablePodsOrPodsWithBadStatusConditions() keep skipping) from "running but not ready", and allow expiry after a bounded wait (e.g. 30m).

internal/controller/humiocluster_controller.go
```diff
@@ -2164,6 +2174,96 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
...
+		return expiredPods[i].CreationTimestamp.Time.Before(expiredPods[j].CreationTimestamp.Time)
+	})
+
+	if *hnp.GetUpdateStrategy().EnableZoneAwareness && !helpers.UseEnvtest() {
+		var zoneFilteredPods []corev1.Pod
+		targetZone := ""
+		for _, pod := range expiredPods {
+			if pod.Spec.NodeName == "" {
+				continue
+			}
```
barkardk, line 2238: Issue: When all expired pods are unscheduled (NodeName == ""), the continue statement skips them entirely, leaving zoneFilteredPods nil. Line 2250 then clears expiredPods,  causing the deletion loop to do nothing. Line 2264 still returns a 1-second requeue,  creating a busy loop that performs zero deletions with no logging to explain the behavior.

Suggestion: Track unscheduled pods separately and return a longer backoff (e.g., 5 seconds) 
with a warning log when unscheduled pods are the reason for the empty filter.

internal/controller/humiocluster_controller.go
```diff
@@ -2164,6 +2174,96 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
...
+			if zone == targetZone {
+				zoneFilteredPods = append(zoneFilteredPods, pod)
+			}
+		}
+		expiredPods = zoneFilteredPods
+	}
+
+	deletionBudget := podsStatus.scaledMaxUnavailableMinusNotReadyDueToMinReadySeconds()
+	for i := 0; i < deletionBudget && i < len(expiredPods); i++ {
```
barkardk, line 2254: When deletionBudget is 0, the loop doesn't execute but line 2264 still returns a 1-second requeue. 
This causes unnecessary reconciliation cycles that do nothing until the budget becomes available.
Recommend: Return a longer backoff (e.g., 30 seconds) when deletionBudget == 0 to reduce wasted cycles.

internal/controller/humiocluster_controller.go
```diff
@@ -2164,6 +2174,96 @@ func (r *HumioClusterReconciler) ensureMismatchedPodsAreDeleted(ctx context.Cont
...
+			if nearestExpiry == 0 || remaining < nearestExpiry {
+				nearestExpiry = remaining
+			}
+		}
+	}
+
+	if len(expiredPods) == 0 {
+		if nearestExpiry > 0 {
+			r.Log.Info(fmt.Sprintf("no expired pods found, next expiry in %s", nearestExpiry))
+			return reconcile.Result{RequeueAfter: nearestExpiry}, nil
```
barkardk, line 2223: Nitpick: 
nearestExpiry prints as e.g. 23h59m59.874392847s — truncate to time.Second. A healthy cluster emits this line every reconcile; consider V(1)/debug level

api/v1alpha1/humiocluster_types.go
```diff
@@ -280,6 +280,12 @@ type HumioNodeSpec struct {
 
 	// PodDisruptionBudget defines the PDB configuration for this node spec
 	PodDisruptionBudget *HumioPodDisruptionBudgetSpec `json:"podDisruptionBudget,omitempty"`
+
+	// ExpireAfter is the maximum duration a pod is allowed to run before being
+	// replaced. When set, pods older than this duration are cycled using the
+	// configured update strategy safety gates (MaxUnavailable, MinReadySeconds,
+	// zone awareness). Uses Go duration format (e.g. "24h", "168h").
+	ExpireAfter *metav1.Duration `json:"expireAfter,omitempty"`
```
barkardk, line 288: No CRD validation. expireAfter: "0s" or a negative value marks every pod instantly expired → churn that can cascade with PDBs. Add kubebuilder markers (Type=string, Pattern=^([1-9][0-9]*)(ns|us|µs|ms|s|m|h)$) and enforce a semantic minimum

